PR #1, opened 2020-08-10
i18n: add Chinese (Traditional) support

vongola12324:
## Description
Add the Traditional Chinese Language (`zh-Hant.js`).
Current Chinese Language is Simplified Chinese, and the file is renamed as `zh-Hans.js` .
Due to js limit, I can not use `'zh-Hant'` in export statement directly, so I have rewrite the export statement in `language.ts` and some code in `app.tsx`.
Also I found some wrong (typo) or missing i18n key-pair in view file and locale file. I added the missing key into `en.js` and fix the typo in view file.

## Screenshots (if appropriate):
截圖 2020-08-10 下午6 24 27

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [ ] All new and existing tests passed.


jesec:
Thanks for your contribution.

Commit 5fb5e771492c59ca7aa8316cebe842aa37b3dc14 is not applicable as "npm install" would discard the change. The package in question "kind-of" is used by dev packages ("webpack", "webpack-dev-server" and "react-dev-utils) only and unfortunately their dependencies locked a lower version of "kind-of" which leads to the "npm audit" warning. The risks to users are extremely low and nothing can be done here unless those packages update their dependencies. 

I squashed and reorganized the commits. 

I tested it with "npm run build; npm run start" and it works fine. But unfortunately it fails if it was deployed to a remote web server. It also fails when you use "npm run start:development:server" and "npm run start:development:client" to simulate server-client environment. 
```
Uncaught TypeError: Cannot read property 'default' of undefined
    at FloodApp.render (home/jc/flood/client/src/javascript/app.tsx:148)
```

vongola12324:
My bad. I did not test in `npm run start:development:server` and `npm run start:development:client` before.
But, that's weird.
I'm now using macOS 10.15.5 with NodeJS 12.18.3 and I have just tested `npm run start:development:server` and `npm run start:development:client`. Neither the browser console nor NodeJS console showed that error. (I can load i18n normally/successfully)

dev:client console
dev:server console
browser console

(The error logs in browser console are about missing i18n text, not the error we focus on)

jesec:
I think a remote web server with rTorrent connected might be necessary to reproduce the problem. I made some changes which appear to solve the issue. Please check it out.

vongola12324:
I have run the code with a rTorrent instance.
It seems that it works correctly without any problems.

vongola12324:
It seems all right. I have no problem with it.

vongola12324:
Thanks.

PR #2, opened 2020-08-11
i18n: automatically detect user's language

jesec:


jesec:
@vongola12324 Please test this.

vongola12324:
I have tested on macOS 10.15 with NodeJS 12.
The client can detect my locale (Traditional Chinese) correctly without any error.

PR #27, opened 2020-08-28
feature: allow to replace main tracker of torrents

jesec:
rTorrent does not have function to remove trackers from torrents
so we only disable existing tracker and insert the new one.

This leads to duplicate listings of trackers in "Trackers" and
"Filter by trackers". We can fix this later by only using enabled
trackers for "Filter by" and "Torrent Details".

Currently this feature only deals with the main tracker (index: 0).
Combined with "Filter by trackers", it is ideal for a bulk change
of tracker URL of private torrents due to a key leak or rotation.

In the future, we can probably make the feature more sophisticated.


jesec:
Probably it should be hidden behind `Advanced` menu by default as it won't be useful for majority of users and can cause problems when triggered improperly.

jesec:
Relevant issues: 
https://github.com/Flood-UI/flood/issues/322
https://github.com/Flood-UI/flood/issues/389
https://github.com/Flood-UI/flood/issues/861

Adding tracker will be much more difficult to manage. But while we are at here, it may be a good idea to take a look at it. 

Deleting tracker is a different story. rTorrent/libTorrent has no API to allow deletion of trackers. ruTorrent directly modifies torrents. It is dangerous and is not a good/stable approach. Let’s see if this is manageable via enable/disable interface. If not, an API change to rTorrent needs to be purposed and a fork might be necessary.

strepito66:
any probability to get a "pastable" set trackers section? 
So that i can paste my trackers without doing it 1 by 1

PR #37, opened 2020-09-16
server: migrate to TypeScript

jesec:


jesec:
I decided to do this in a incremental way. New changes will be pushed to master. Thus, PR won't be necessary.

Issue #44, opened 2020-10-03
SyntaxError: Unexpected token '.'


artik:
Hi, I'm having a new error when launching Flood in a docker container. I tried the latest commit: `bdb8144f4b1d4921d9d60f28424aab5c31f455f6`. Here is the error:

```
SyntaxError: Unexpected token '.'
    at Module._compile (internal/modules/cjs/loader.js:891:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10)
    at internal/main/run_main_module.js:17:11
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! flood@3.1.0 start: `node --use_strict dist/index.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the flood@3.1.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Here is my Dockerfile:

```
FROM alpine:3.11

ARG RTORRENT_VER=0.9.8
ARG LIBTORRENT_VER=0.13.8
ARG MEDIAINFO_VER=19.09
ARG BUILD_CORES

ENV UID=1024 GID=100 \
    FLOOD_SECRET=supersecret \
    WEBROOT=/ \
    RTORRENT_SCGI=0 \
    PKG_CONFIG_PATH=/usr/local/lib/pkgconfig

RUN NB_CORES=${BUILD_CORES-`getconf _NPROCESSORS_CONF`} \
 && apk -U upgrade \
 && apk add -t build-dependencies \
    build-base \
    git \
    libtool \
    automake \
    autoconf \
    wget \
    tar \
    xz \
    zlib-dev \
    cppunit-dev \
    openssl-dev \
    ncurses-dev \
    curl-dev \
    binutils \
    linux-headers \
 && apk add \
    ca-certificates \
    curl \
    ncurses \
    openssl \
    gzip \
    zip \
    zlib \
    s6 \
    su-exec \
    python \
    nodejs \
    nodejs-npm \
    unrar \
    findutils \
 && cd /tmp && mkdir libtorrent rtorrent \
[...]

PR #46, opened 2020-10-04
build: fix target version to support Node 12 & up

Roguyt:
## Description
While the README says flood only support the current version of NodeJS, the package.json indicates that NodeJS >= 12 is supposed to work.
This PR is a fix on the TS config to allow the transpiled code to run on version that doesn't support optional chaining. Cf https://github.com/jesec/flood/issues/44#issuecomment-703281963

The fix is based on Typescript documentation regarding the support of target and targeted NodeJS version: https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping

## Related Issue
As explained in this issue Flood does compile on Node 12 but crashes at runtime because of the optional chaining.
In order to reproduce:
1. Use Node 12
2. Clone the repo
3. Build from the source
4. Launch Flood

## Motivation and Context
While supporting the last version is a good thing, supporting at least the current LTS version is something that isn't too hard and could help in the case of the stability that gives LTS version.

## How Has This Been Tested?
This is just a change of the compiler option, which doesn't affect the functionality or code at all of the project.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.

### DIFF
--- a/server/tsconfig.json
+++ b/server/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
+    "target": "es2019",
     "module": "commonjs",
     "noEmit": false,
     "sourceMap": false,


jesec:
Client side is managed by [browserslist](https://github.com/browserslist/browserslist) so I updated the PR to change target version of server only.

PR #53, opened 2020-10-10
auth, Users: initial preparation for multi client support

jesec:
~BREAKING CHANGE.~

~Requires database reset.~ https://github.com/jesec/flood/pull/53/commits/8c3e9eccfb75f75eebc06daf21f1452d391b254b.

Bug: Flood-UI/Flood#185

<details>
    <summary>screenshot (946a972):</summary>

image

</details>

Data structures defined in: https://github.com/jesec/flood/tree/refactor/auth/shared/schema


jesec:
@zawapete This is quite a large change to the user management. I will be appreciated if you, as original author of the user management feature, can give some insights.

aspacca:
@jesec I will take a look it

zawapete:
@jesec I went throught the commit. I can try to test on my machine with multiple users but the db I have is from original flood repo codebase (even a little old). do I need to run some kind of migration before switching to this repo code (and then this branch)?

jesec:
> @jesec I went throught the commit. I can try to test on my machine with multiple users but the db I have is from original flood repo codebase (even a little old). do I need to run some kind of migration before switching to this repo code (and then this branch)?

I implemented the migration in https://github.com/jesec/flood/commit/8c3e9eccfb75f75eebc06daf21f1452d391b254b. 

If you came from original Flood repo and if you use the new command line config interface, you probably have to pay attention to the runtime directory. It is now in `~/.local/share/flood` by default instead of directory of Flood's sources (so system-wide installation to privileged directories like `/usr/bin` can work). There might be other breaking changes there, for example minimum Node version and handling of baseURI. You have to check the changelog and make appropriate changes. 

I would recommend just delete the old Flood and start fresh if you want to skip the hurdles.

zawapete:
> It is now in `~/.local/share/flood` by default instead of directory of Flood's sources

that's why moving the old db didn't work :)

oki, I'm able to run flood from this repo, migrating the old databse etc. but `GET /flood/api/directory-list` returns 500 (both with allowedpath or without). any hint?

jesec:
> > It is now in `~/.local/share/flood` by default instead of directory of Flood's sources
> 
> that's why moving the old db didn't work :)
> 
> oki, I'm able to run flood from this repo, migrating the old databse etc. but `GET /flood/api/directory-list` returns 500 (both with allowedpath or without). any hint?

When and how does it happen? 

Check response data. It probably contains the error message. 

Alternatively run a debugger and break at https://github.com/jesec/flood/blob/master/server/routes/api/index.ts#L33, https://github.com/jesec/flood/blob/master/server/util/fileUtil.ts#L45.

zawapete:
I was able to create a user once logged in as a non admin user.

zawapete:
I cannot run a debbugger unlucily on the instance I'm testing.

response from the route is an empty object: `{}`

> When and how does it happen?

when I try to add a torrent and select the path where to save
or when I try to set the location of an existinng torrent

jesec:
> I was able to create a user once logged in as a non admin user.

This case is covered and validated by https://github.com/jesec/flood/blob/de83b38b19a4599e585066e2912ba11b4c071825/server/routes/api/auth.test.ts#L130. It should not happen. 

I suspect there is something wrong with migration. Old Flood doesn't have types and as a result some data structures are messed up (eg. should be a boolean but got string 'true' / 'false' which JS will probably evaluate to boolean true). 

It is handled in: https://github.com/jesec/flood/blob/7ca397b0743dcb729acfbbe0c5e23fa2f9827f65/server/bin/migrations/UserInDatabase2.ts#L61.

jesec:
> I cannot run a debbugger unlucily on the instance I'm testing.

Try VS Code remote. It should work as long as you have SSH access.

> 
> response from the route is an empty object: `{}`
> 

That's interesting. Technically that won't happen. No code path there throw an empty error. 

> > When and how does it happen?
> 
> when I try to add a torrent and select the path where to save
> or when I try to set the location of an existinng torrent

Can you try to type in an allowed and existing path? 

BTW: Note that when you use `npm start`, you have to type arguments after `--`, for example `npm start -- --allowedpath /some/path`.

zawapete:
> I suspect there is something wrong with migration. Old Flood doesn't have types and as a result some data structures are messed up (eg. should be a boolean but got string 'true' / 'false' which JS will probably evaluate to boolean true).

I created the not admin user from master, but it was created as admin

from this branch existing users where migrated properly and newly non admin users as well

I will try now to debug the directory-list issue

jesec:
> > I suspect there is something wrong with migration. Old Flood doesn't have types and as a result some data structures are messed up (eg. should be a boolean but got string 'true' / 'false' which JS will probably evaluate to boolean true).
> 
> I created the not admin user from master, but it was created as admin
> 

The issue is introduced by https://github.com/jesec/flood/commit/f6896413a3979d6235f5538ac8263d007b9ac88d which deduplicated `/register` and `PUT /users/{username}` endpoints. `/register` endpoint uses `isAdmin: true` unconditionally while `create-user` endpoint takes `isAdmin` into account. It is (unintentionally) fixed by https://github.com/jesec/flood/pull/53/commits/cdfe164abccf8be05906e73ba25a79fc18bd4d73#diff-d5df166a5f0790c6486e950129e7775bR167 . I guess I probably need to do a separate fix for master branch if this PR is still pending. 

> from this branch existing users where migrated properly and newly non admin users as well
> 

Good to know.

zawapete:
```
null: TypeError: (inputPath || "/").replace is not a function
    at Object.exports.getDirectoryList (/opt/flood/server/util/fileUtil.ts:46:41)
    at /opt/flood/server/routes/api/index.ts:35:3
    at Layer.handle [as handle_request] (/opt/flood/node_modules/express/lib/router/layer.js:95:5)
    at next (/opt/flood/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/opt/flood/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/opt/flood/node_modules/express/lib/router/layer.js:95:5)
    at /opt/flood/node_modules/express/lib/router/index.js:281:22
    at Function.process_params (/opt/flood/node_modules/express/lib/router/index.js:335:12)
    at next (/opt/flood/node_modules/express/lib/router/index.js:275:10)
    at exports.default (/opt/flood/server/middleware/appendUserServices.ts:18:3) {stack: 'TypeError: (inputPath || "/").replace is not …server/middleware/appendUserServices.ts:18:3)', message: '(inputPath || "/").replace is not a function'}
```

jesec:
> ```
> null: TypeError: (inputPath || "/").replace is not a function
>     at Object.exports.getDirectoryList (/opt/flood/server/util/fileUtil.ts:46:41)
>     at /opt/flood/server/routes/api/index.ts:35:3
>     at Layer.handle [as handle_request] (/opt/flood/node_modules/express/lib/router/layer.js:95:5)
>     at next (/opt/flood/node_modules/express/lib/router/route.js:137:13)
>     at Route.dispatch (/opt/flood/node_modules/express/lib/router/route.js:112:3)
>     at Layer.handle [as handle_request] (/opt/flood/node_modules/express/lib/router/layer.js:95:5)
>     at /opt/flood/node_modules/express/lib/router/index.js:281:22
>     at Function.process_params (/opt/flood/node_modules/express/lib/router/index.js:335:12)
>     at next (/opt/flood/node_modules/express/lib/router/index.js:275:10)
>     at exports.default (/opt/flood/server/middleware/appendUserServices.ts:18:3) {stack: 'TypeError: (inputPath || "/").replace is not …server/middleware/appendUserServices.ts:18:3)', message: '(inputPath || "/").replace is not a function'}
> ```

Fixed by 2259ee18d1de896908e5c1a9a17eb93da9e166d0 in `master`.

zawapete:
so I'll wait you merge/rebase this branch from master?

jesec:
> so I'll wait you merge/rebase this branch from master?

It is done.

zawapete:
I have now the error message `This path does not exist. It will be created.` instead of empty message but the underlying problem seems still present.
I see that you return that message if `typeof inputPath !== 'string'`, but the problem is that `inputPath` is not a string for some reason: either but typing a path or using the default already set

zawapete:
https://github.com/jesec/flood/blob/master/server/routes/api/index.ts#L35
```
  getDirectoryList(req.path)
```

seems to fix the issue: I still have problem accessing the folder anyway.
If I set allowedpaths the error message it says that I don't have acces. If I don't set them (correct me if I'm wrong: this will give access to full filesystem) it says that the path doesn't exists and has to be created

zawapete:
no, sorry, my mistake

zawapete:
```
 getDirectoryList(req.query.path)
```

jesec:
Indeed, there is an internal API mismatch. 

Client action (https://github.com/jesec/flood/blob/master/client/src/javascript/actions/FloodActions.ts#L134) does not match server API endpoint (https://github.com/jesec/flood/blob/master/server/routes/api/index.ts#L33). 

I have fixed this in commit 6217dbd324f0921ea (included in #48). But due to slow actions in DefinitelyTyped, I have not merged the PR.

jesec:
The change is ported to `master` in https://github.com/jesec/flood/commit/a25ed1742268b6bbfea41a39c54ae33e9c34fcc0.

Issue #104, opened 2020-11-22
Add docker runtime configuration for flood process UID and GID


r0bj:
Type: Feature Request

- [x] If you want to contribute to the project please review the [contributing guidelines](https://github.com/Flood-UI/flood/blob/master/.github/CONTRIBUTING.md).

## Summary

Flood heavily rely on filesystem permissions for accessing files which are shared between other processes (e.g. rtorrent). Currently flood process UID and GID are assigned during image building and are 1001. So in order to allow flood to access files we need to change permissions to less secure model (add access for all users) which is not perfect.

## Idea of implementation

Allow to set UID and GID at container runtime via e.g. environment variables.

r0bj:
Close it with comment: https://github.com/jesec/flood/pull/105#issuecomment-731698244

PR #105, opened 2020-11-22
Add ability to set UID and GID during container runtime

r0bj:
Add ability to set UID and GID during container runtime via env variables.

## Description

Currently user `download` which is used as a flood process owner is created at image build. It doesn't give much flexibility if we would like to change process UID and GID. Changing those is useful for flood because it heavily rely on filesystem permissions for accessing files which are shared between other processes (rtorrent).
This PR changes the way process owner user is created, instead of creating user at build time it's created at container runtime. This way we can start container with desired UID and GID, e.g.:
```
docker run -it -e FLOOD_UID=2000 -e FLOOD_GID=2000 jesec/flood
```

Setting those env variables is optional.

## Related Issue

Fixes https://github.com/jesec/flood/issues/104

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have read the **CONTRIBUTING** document.
- [x] All new and existing tests passed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,16 +37,13 @@ RUN rm /tmp/*.tgz
 
 # Install runtime dependencies
 RUN apk --no-cache add \
-    mediainfo
-
-# Create "download" user
-RUN adduser -h /home/download -s /sbin/nologin --disabled-password download
-
-# Run as "download" user
-USER download
+    mediainfo su-exec
 
 # Expose port 3000
 EXPOSE 3000
 
+# Copy docker entrypoint script
+COPY docker-entrypoint.sh /
+
 # Flood
-ENTRYPOINT ["flood", "--host=0.0.0.0"]
+ENTRYPOINT ["/docker-entrypoint.sh", "flood", "--host=0.0.0.0"]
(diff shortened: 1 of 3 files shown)


jesec:
I am reluctant about this. It is a wide-known security issue to run with higher privileges first and then reduce the privileges in the same TTY context (https://github.com/ncopa/su-exec/issues/23). 

It might be more appropriate to use uid mapping (namespace) feature of Docker. 

Alternatively, given Flood docker image is simply composed with `sudo npm install --global flood`, you may:

- Simply use `--user` argument of `docker run`. Flood is globally installed inside the Docker image and it can be executed by any user inside the container as long as there is a readable/writeable `rundir`. Flood doesn't actually need a home directory. The image creates a user simply because the default runtime directory is `$HOME/.local/share/flood`. In practice, if you want the settings to be persistent, `rundir` is always a mapped folder. 
- Compose a Docker image with appropriate UID/GID. It is extremely easy to do so. `docker build --tag flood https://github.com/jesec/flood/raw/master/Dockerfile.release` If you think this solution suits you, you can create a separate PR that makes version, uid and gid configurable in the Dockerfile if necessary. 
- Or, just install Node.js and run `sudo npm install --global flood` in your own configured Docker image. It is that simple. Official Flood docker image is not that special, really.

r0bj:
Yes, you are right, it's easy enough to build custom flood image. Thanks!
Let's close PR then.

PR #182, opened 2021-02-01
Add role attribute to clickable elements

trailofdead:
Add support for vimium keyboard navigation and screen readers

  * Add menuitem role to SidebarFilter li
  * Add button role to Tooltip div

## Description
Adds `role="menuitem"` to clickable `li` elements in SidebarFilter and `role="button" tabIndex={0}` to `div` elements in Tooltip

## Related Issue
https://github.com/Flood-UI/flood/issues/885

## Screenshots
tooltip
sidebar

## Types of changes

### DIFF
--- a/client/src/javascript/components/general/Tooltip.tsx
+++ b/client/src/javascript/components/general/Tooltip.tsx
@@ -402,6 +402,8 @@ class Tooltip extends Component<TooltipProps, TooltipStates> {
     return (
       <div
         className={wrapperClassName}
+        role="button"
+        tabIndex={0}
         onClick={onClick}
         onMouseEnter={() => this.handleMouseEnter()}
         onMouseLeave={() => this.handleMouseLeave()}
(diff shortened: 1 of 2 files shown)


jesec:
This change seems to introduce some visual border when the element is focused. It does not align with existing designs.

trailofdead:
The issue you're describing is caused because the div is now focusable, since the addition of the `tabIndex` attribute.  This project uses the eslint jsx-a11y/interactive-supports-focus rule, and removing tabIndex makes the div unfocusable and violates the rule, hurting accessibility.

jesec:
Visual regressions are not preferable, either. I am going to rollback the tabIndex for now unless there is a better solution.

trailofdead:
Can we keep the role attribute, remove tabIndex, and ignore the eslint rule for that line?

jesec:
I fixed the visual regression while preserving the a11y feature with 9f21f943c5c712ad8fdc0049f67d6edcd1bc9854 and 973de1ac6b2a6526d7a8cbe61e7daebf871e55fe.

PR #220, opened 2021-02-28
rTorrent: use JSON-RPC when available

jesec:
Preliminary tests: (compared to XMLRPC)

2x performance, 15% lower total CPU time in rTorrent process, 33% lower total CPU time in Node.js process. Note that there are 28000+ entries, so it is natural to take seconds to complete the request. Additionally, current implementation on the rTorrent side uses nlohmann's `JSON for Modern C++`, which is the easiest to use but it is not one of the most "High Performance" JSON library. Some JSON libraries feature up to [8x performance](https://github.com/miloyip/nativejson-benchmark#stringify-time) over nlohmann/json. 

image

### DIFF
--- a/server/services/rTorrent/clientGatewayService.ts
+++ b/server/services/rTorrent/clientGatewayService.ts
@@ -804,14 +804,29 @@ class RTorrentClientGatewayService extends ClientGatewayService {
     torrentTracker: string[];
     transferSummary: string[];
   }> {
-    const methodList: Array<string> = await this.clientRequestManager
-      .methodCall('system.listMethods', [])
-      .then(this.processClientRequestSuccess, this.processRTorrentRequestError)
-      .catch((e) => {
+    let methodList: Array<string> = [];
+    const listMethods = () => {
+      return this.clientRequestManager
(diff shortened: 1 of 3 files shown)


sabersalv:
@jesec Does the [jesc/rtorrent:latest](https://hub.docker.com/layers/jesec/rtorrent/latest/images/sha256-e769e8d433709478ae4e819dddc2548f5902c9fee46c49872982e8d237e30096?context=explore) docker image have JSON RPC enabled? How can I verify the communication is using JSON RPC over XLM PRC？

jesec:
> @jesec Does the [jesc/rtorrent:latest](https://hub.docker.com/layers/jesec/rtorrent/latest/images/sha256-e769e8d433709478ae4e819dddc2548f5902c9fee46c49872982e8d237e30096?context=explore) docker image have JSON RPC enabled? How can I verify the communication is using JSON RPC over XLMPRC？

Yes.

Unfortunately, Flood does not print out that information. If you do want to verify that, you would have to start a debugging session and inspect the code paths. You can also try to fix `isJSONCapable` to true, compile Flood and see if things work.

PR #228, opened 2021-03-07
feat: api authentication support Authorization header

sabersalv:
## Description

[Browsers block frontend JavaScript code from accessing the `Set Cookie` header, as required by the Fetch spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie), so the current cookie method is not friendly for web fetch.

This PR allows API to use the Bearer header for authentication. 

If you are interested in this PR, I can add unit tests.

## Test plan

1. call `/auth/authenticate` returns `{ token }`
2. call `/client/connection-test` with `Authorization: Bearer <token>` header
3. response success

- [x] follow the test plan

## Types of changes

### DIFF
--- a/server/config/passport.ts
+++ b/server/config/passport.ts
@@ -1,4 +1,4 @@
-import {Strategy} from 'passport-jwt';
+import {Strategy,ExtractJwt} from 'passport-jwt';
 
 import type {PassportStatic} from 'passport';
 import type {Request} from 'express';
@@ -11,7 +11,7 @@ import Users from '../models/Users';
 // Setup work and export for the JWT passport strategy.
 export default (passport: PassportStatic) => {
   const options = {
-    jwtFromRequest: (req: Request) => req?.cookies?.jwt,
+    jwtFromRequest: (req: Request) => req?.cookies?.jwt || ExtractJwt.fromAuthHeaderAsBearerToken()(req),
     secretOrKey: config.secret,
   };
 
--- a/server/config/passport.ts
+++ b/server/config/passport.ts
@@ -1,4 +1,4 @@
-import {Strategy,ExtractJwt} from 'passport-jwt';
+import {Strategy, ExtractJwt} from 'passport-jwt';
 
 import type {PassportStatic} from 'passport';
 import type {Request} from 'express';
--- a/server/routes/api/auth.ts
+++ b/server/routes/api/auth.ts
@@ -45,12 +45,14 @@ const sendAuthenticationResponse = (
 ): Response => {
   const {username, level} = credentials;
 
-  res.cookie('jwt', getAuthToken(username), getCookieOptions());
+  const token = getAuthToken(username);
+  res.cookie('jwt', token, getCookieOptions());
 
   const response: AuthAuthenticationResponse = {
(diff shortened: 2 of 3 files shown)


jesec:
If a web API has such intentional limitation, there are reasons. 

I am concerned about the security implications of exposing the token to JavaScript VM, as it might be possible for some untrusted scripts from Internet to exploit user's browser to get a token (and send it back to attacker). 

Can you try to set appropriate parameters for fetch's `credentials` option?

sabersalv:
I'm working on a web application that consumes Flood API, and hit this limit. I think this PR is secure, as untrusted scripts won't be able to intercept the request and read the token from the response or VM memory (if it can, it probably can read the cookie too)

I did some tests with your suggestion `credentials` option, it does not work.

1. The app will get credentials from the user and do the authentication by calling `fetch('https://flood.com/api/auth/authenticate)`, the browser won't store the cookie. It breaks the authentication chain.
2. Let's assume the cookie is already stored under the `flood.com` domain, call `fetch` with `{ credentials: 'include' }` won't send the cookie back to the server, [Reason: Credential is not supported if the CORS header ‘Access-Control-Allow-Origin’ is ‘*’](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials)

In summary, looks like there's no way for a pure web application to do the authentication with Flood API.

Luckily, mine is an Electron web application and I figured a way out by using `node-fetch` to read the cookie. I'm going to close this one.

jesec:
> I'm working on a web application that consumes Flood API, and hit this limit. I think this PR is secure, as untrusted scripts won't be able to intercept the request and read the token from the response or VM memory (if it can, it probably can read the cookie too)
> 
> I did some tests with your suggestion `credentials` option, it does not work.
> 
> 1. The app will get credentials from the user and do the authentication by calling `fetch('https://cross-origin.com/api/auth/authenticate)`, but the browser won't store the cookie. It breaks the authentication chain in the first place.
> 2. Let's assume the cookie is already stored under the `cross-origin.com` domain, call `fetch` with `{ credentials: 'include' }` won't send the cookie back to the server, [Reason: Credential is not supported if the CORS header ‘Access-Control-Allow-Origin’ is ‘*’](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials)
> 
> In summary, looks like there's no way for a pure web application to do the authentication with Flood API using the cookie method.
> 
> Luckily, mine is an Electron web application and I figured a way out by using `node-fetch` to get the cookie. I'm going to close this one.

Flood sets cookies with `HttpOnly` attribute. This specifically prevents scripts from accessing token directly and mandates that a script running in browser must comply with browser's security mechanisms. 

It is by design that a script from other (potentially untrusted) sites can't access Flood with stored credentials.

However, it remains possible to interact with Flood via a web application. For example, browser plugins are exempted from cross-site constraints. Additionally, as you said, local applications like Electron or Node.js script, can interact with Flood without the browser as a gatekeeper. 

IIRC you probably have to use `include` for both authentication and subsequent requests.

PR #235, opened 2021-03-09
Include rtorrent output in log

rasmusbe:
## Description

Include rtorrent output in the log.

## Types of changes


jesec:
This is unnecessary. `stdio` is already `inherit`.

rasmusbe:
You're probably right, I was hoping this would solve that print-commands in rtorrent.rc would work, but it seems to be the same with this fix.

PR #237, opened 2021-03-10
server: qBittorrent: Update the status filter behavior

slikie:
server: qBittorrent: Update the status filter behavior so that UX is in sync with qB GUI/WebUI.

## Description

Mainly appends 'pausedDL' torrents to 'Downloading' status.

I understand rtorrent/rutorrent may have the design of classifying torrent status, and I'm not sure if you want to unify the logic for the current three backends, but this PR would be more natural logically for qBittorrent users.

Adjustment based on:
https://github.com/CzBiX/qb-web/blob/809d5bab4b5b4566cd3d74eb6e5df243fb4a17c1/src/utils.ts#L4

## Types of changes

### DIFF
--- a/server/services/qBittorrent/util/torrentPropertiesUtil.ts
+++ b/server/services/qBittorrent/util/torrentPropertiesUtil.ts
@@ -68,6 +68,7 @@ export const getTorrentStatusFromState = (state: QBittorrentTorrentState): Torre
       statuses.push('downloading');
       break;
     case 'pausedDL':
+      statuses.push('paused');
       statuses.push('inactive');
       statuses.push('stopped');
       break;
--- a/client/src/javascript/components/sidebar/StatusFilters.tsx
+++ b/client/src/javascript/components/sidebar/StatusFilters.tsx
@@ -29,6 +29,11 @@ const StatusFilters: FC = observer(() => {
       icon: <DownloadSmall />,
     },
     {
+      label: i18n._('filter.status.paused'),
+      slug: 'paused',
+      icon: <UploadSmall />,
+    },
+    {
       label: i18n._('filter.status.seeding'),
       slug: 'seeding',
       icon: <UploadSmall />,
--- a/client/src/javascript/i18n/strings/en.json
+++ b/client/src/javascript/i18n/strings/en.json
@@ -131,6 +131,7 @@
   "filter.status.checking": "Checking",
   "filter.status.completed": "Complete",
   "filter.status.downloading": "Downloading",
+  "filter.status.paused": "Paused",
   "filter.status.error": "Error",
   "filter.status.inactive": "Inactive",
   "filter.status.seeding": "Seeding",
--- a/client/src/javascript/util/torrentStatusClasses.ts
+++ b/client/src/javascript/util/torrentStatusClasses.ts
@@ -10,6 +10,7 @@ function torrentStatusClasses(
     'torrent--has-error': status.includes('error'),
     'torrent--is-stopped': status.includes('stopped'),
     'torrent--is-downloading': status.includes('downloading'),
+    'torrent--is-paused':  status.includes('paused'),
     'torrent--is-downloading--actively': downRate > 0,
     'torrent--is-uploading--actively': upRate > 0,
     'torrent--is-seeding': status.includes('seeding'),
--- a/shared/constants/torrentStatusMap.ts
+++ b/shared/constants/torrentStatusMap.ts
@@ -1,6 +1,7 @@
 const torrentStatusMap = [
   'checking',
   'seeding',
+  'paused',
   'complete',
   'downloading',
   'stopped',


jesec:
None of the changes is effective. Note that switch statements fallthroughs till `break` and a torrent can’t be `downloading` and `stopped` at the same time.

slikie:
The main reason for this PR is that there is currently no apparent way to filter `pausedDL` torrents for continuing in the current category. I understand that it would be confusing that a torrent is both in `downloading` and `stopped`. Do you think it's an idea to rename `stopped` to `paused` like qB? 

I only mention ruTorrent because it was de-facto the main rTorrent webui before Flood so maybe there is some experience or design to inherit. Disregard if you don't care

jesec:
> The main reason for this PR is that there is currently no apparent way to filter `pausedDL` torrents for continuing in the current category. I understand that it would be confusing that a torrent is both in `downloading` and `stopped`. Do you think it's an idea to rename `stopped` to `paused` like qB?

I don't think that's necessary. IMO there is zero reason to make "stopped" distinct from "paused". It only creates more confusion.

jesec:
@slikie 

Hi, are you still championing this PR?

If yes, you would have to at least make it actually effective. (and respond to my comments)

if no, feel free to close the PR.

slikie:
> and a torrent can’t be `downloading` and `stopped` at the same time.
I don't think there's any code logic present against that.

> None of the changes is effective. Note that switch statements fallthroughs till `break` 
I tested it and it is effective. One torrent is only at a state at the time, it won't need to go past `break`.

I'm still arguing that the current UX doesn't provide a easy way to **resume previously paused torrents that aren't finished downloading**, which is a necessary function.

There are a lot of circumstances people need to pause their torrents and resume afterwards, and there isn't an apparent queue in flood right now to address that.

Right now I will need to go `stopped` status, sort by percent complete, and  **by no way I can tell if one torrent is finished downloading or stuck at 99%**.

Hope that address your confusion. 

There are a few ways for this:

- Transmission/Deluge: add `paused` and refactor `stopped`  into `completed`.
- rutorrent: add `pausedDL` into `Downloading`, as I proposed. while there is no `stopped` category.
- qBittorrent: add `pausedDL` into `Downloading` and refactor `stopped` into `Paused`.

jesec:
You don’t need to change any backend code for this. 

What you want is a filter that contains “stopped” but not “completed”.

server/services/qBittorrent/util/torrentPropertiesUtil.ts
```diff
@@ -62,16 +62,22 @@ export const getTorrentStatusFromState = (state: QBittorrentTorrentState): Torre
...
+      statuses.push('downloading');
+      break;
     case 'downloading':
     case 'forcedDL':
       statuses.push('active');
       statuses.push('downloading');
       break;
     case 'pausedDL':
       statuses.push('inactive');
+      statuses.push('downloading');
```
jesec, line 74: This is the only effective line.

server/services/qBittorrent/util/torrentPropertiesUtil.ts
```diff
@@ -62,16 +62,22 @@ export const getTorrentStatusFromState = (state: QBittorrentTorrentState): Torre
       statuses.push('downloading');
       break;
     case 'metaDL':
+      statuses.push('downloading');
+      break;
     case 'downloading':
     case 'forcedDL':
       statuses.push('active');
```
jesec, line 69: `metaDL` previously goes to here. Your change only removes `active` from the status. Is that intended? `metaDL` should be an active status.

server/services/qBittorrent/util/torrentPropertiesUtil.ts
```diff
@@ -62,16 +62,22 @@ export const getTorrentStatusFromState = (state: QBittorrentTorrentState): Torre
...
       statuses.push('inactive');
+      statuses.push('downloading');
       statuses.push('stopped');
       break;
     case 'queuedDL':
+      statuses.push('inactive');
+      statuses.push('downloading');
+      break;
     case 'stalledDL':
       statuses.push('inactive');
```
jesec, line 82: `queuedDL` previously goes to here. Your change is not effective.

jesec:
> You don’t need to change any backend code for this.
> 
> What you want is a filter that contains “stopped” but not “completed”.

This. 

Current change is unnecessary and simply won’t work. There will not magically be a new `paused` category.

slikie:
> IMO there is zero reason to make "stopped" distinct from "paused". It only creates more confusion.

I was kind of worried about this but I take this as a go-ahead?

jesec:
The commit changes the public API, and as a result, unfortunately, is a breaking change. Per versioning conventions, I can’t do this without a major release.

And, as i said, backend change is totally unnecessary. You can achieve this purely in frontend by filtering “stopped” and “NOT complete”. 

It is an option to add a “Paused” filter that does that in the frontend. However, I would prefer a more general method like “Advanced Filtering” (see Flood-UI/flood#312, though, I can implement this myself in the future).

trim21:
stalled, if you still want this feature you can re-open a new one.

PR #253, opened 2021-03-22
client: add CMD-A, CTRL-A to select all torrents

sabersalv:
## Description

After the user selected a torrent, user can use `CMD-A` or `CTRL-A` to select all torrents. Click one torrent to deselect all. Very useful for bulk editing large collection of torrents.

For https://github.com/jesec/flood/discussions/122#discussioncomment-480117

## Types of changes

### DIFF
--- a/client/src/javascript/components/torrent-list/TorrentList.tsx
+++ b/client/src/javascript/components/torrent-list/TorrentList.tsx
@@ -1,7 +1,8 @@
-import {FC, ReactNode, useEffect, useRef} from 'react';
+import {FC, KeyboardEvent, ReactNode, useEffect, useRef} from 'react';
 import {observer} from 'mobx-react';
 import {reaction} from 'mobx';
 import {Trans} from '@lingui/react';
+import {useEvent} from 'react-use';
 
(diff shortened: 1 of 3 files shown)


jesec:
Good idea. 

I move the event handler to the list component, and attach it to the entire window. I think that would make it more consistent. 

The PR works fine for me so far. Please test and let me know when it is good to go.

client/src/javascript/components/torrent-list/TorrentList.tsx
```diff
@@ -47,6 +49,15 @@ const TorrentList: FC = observer(() => {
     return dispose;
   }, []);
 
+  useEvent('keydown', (e: KeyboardEvent) => {
+    const {repeat, metaKey, ctrlKey, key} = e;
+    if (repeat) return;
```
sabersalv, line 54:
```
 const { target } = e;
 if (['INPUT', 'TEXTAREA'].includes(target.tagName)) return;
```

Can still select all text on `<input>` and `<textarea>`

sabersalv:
Tested, works great. Left one comment.

PR #258, opened 2021-03-29
Add dateAdded based on torrent directory when not present elsewhere

hibes:
I added a feature which provides a `dateAdded` value from the directory's creation date when it's not found in the `torrentProperties` response.

## Description

This is really just an inspirational PR. Curious if you have any ideas for how to implement it better or something. I can never get rtorrent to generate `dateAdded` consistently and correctly so that it can be used by `flood`. Like, some of my torrents are able to retain that field, but some others can't.

## Types of changes

### DIFF
--- a/server/services/rTorrent/clientGatewayService.ts
+++ b/server/services/rTorrent/clientGatewayService.ts
@@ -56,6 +56,8 @@ import {
 import type {MultiMethodCalls} from './util/rTorrentMethodCallUtil';
 import type {RPCError} from './types/RPCError';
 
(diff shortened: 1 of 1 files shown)


jesec:
Filesystem ops are very expensive, and I don't think it is a good idea to do that in `fetchTorrentList`. 

Additionally, the cache will not survive a reboot, and this could lead to a significant I/O spike or even halt when users have many torrents. 

I don't think the change is necessary, either. If users manage their instance with Flood, the `dateAdded` field will be there.

PR #260, opened 2021-03-31
feat: add tini to container images

onedr0p:
## Description

Small change here. I created an `ENTRYPOINT` with [tini](https://github.com/krallin/tini) and updated the `CMD` in the `Dockerfile`s. tini is already included in the `Docker` CRI, however some people (like me) use a different CRI like `containerd` or `cri-o` where we don't have access to an `--init` flag. 

`tini` also has other benefits as described in this [Github issue](https://github.com/krallin/tini/issues/8#issuecomment-146135930).

## Types of changes

Container image change

### DIFF
--- a/distribution/containers/Dockerfile.release
+++ b/distribution/containers/Dockerfile.release
@@ -12,7 +12,8 @@ RUN npm i -g "${PACKAGE}"@"${VERSION}" &&\
 
 # Install runtime dependencies
 RUN apk --no-cache add \
-    mediainfo
+    mediainfo \
+    tini
 
 # Create "download" user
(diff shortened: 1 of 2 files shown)


jesec:
I don't see a strong reason for this change at the moment. However, it won't hurt to have tini, so let's try it out.

jesec:
BTW: Can you apply the change to rolling images, as well? So people can test it.

onedr0p:
I hear yea, but tini is such an incredible tiny piece of software for containers. I'm running rtorrent-flood in Kubernetes with containerd, no Docker :)

PR #261, opened 2021-04-01
fix: flood entrypoint in container images

onedr0p:
## Description

I did not think of the ramifications of moving flood from the entrypoint to the command. Keeping it in the entrypoint should mean it will work the way the documentation expects it too.

Let me know if you want me to move the other flags to the entrypoint as well or if they are expected to be defaults we could leave them out.

 Sorry!

## Types of changes

### DIFF
--- a/distribution/containers/Dockerfile.release
+++ b/distribution/containers/Dockerfile.release
@@ -25,5 +25,5 @@ USER download
 EXPOSE 3000
 
 # Flood
-ENTRYPOINT ["/sbin/tini", "--"]
-CMD ["flood", "--host=0.0.0.0"]
+ENTRYPOINT ["/sbin/tini", "--", "flood"]
+CMD ["--host=0.0.0.0"]
--- a/distribution/containers/Dockerfile.release
+++ b/distribution/containers/Dockerfile.release
@@ -25,5 +25,4 @@ USER download
 EXPOSE 3000
 
 # Flood
-ENTRYPOINT ["/sbin/tini", "--", "flood"]
-CMD ["--host=0.0.0.0"]
+ENTRYPOINT ["/sbin/tini", "--", "flood", "--host=0.0.0.0"]
--- a/distribution/containers/Dockerfile.rtorrent
+++ b/distribution/containers/Dockerfile.rtorrent
@@ -9,5 +9,4 @@ FROM jesec/flood:$FLOOD_VERSION as flood
 COPY --from=rtorrent / /
 
 # Flood with managed rTorrent daemon
-ENTRYPOINT ["/sbin/tini", "--", "flood"]
-CMD ["--host=0.0.0.0", "--rtorrent"]
+ENTRYPOINT ["/sbin/tini", "--", "flood", "--host=0.0.0.0", "--rtorrent"]
--- a/distribution/containers/Dockerfile.rtorrent
+++ b/distribution/containers/Dockerfile.rtorrent
@@ -9,5 +9,5 @@ FROM jesec/flood:$FLOOD_VERSION as flood
 COPY --from=rtorrent / /
 
 # Flood with managed rTorrent daemon
-ENTRYPOINT ["/sbin/tini", "--"]
-CMD ["flood", "--host=0.0.0.0", "--rtorrent"]
+ENTRYPOINT ["/sbin/tini", "--", "flood"]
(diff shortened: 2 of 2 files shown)


jesec:
um.. how about the default CLI arguments?

onedr0p:
I see need for --rtorrent but is --host needed? Does/can the app default to 0.0.0.0?

jesec:
It would be insecure if the app listens without limit by default.

onedr0p:
Limit?

jesec:
> Limit?

i.e. 0.0.0.0.

jesec:
I would like to keep all default arguments.

onedr0p:
Ok I'll update soon.

Why isn't `host` set to `0.0.0.0` in the default in flood settings?

jesec:
> Ok I'll update soon.
> 
> Why isn't `host` set to `0.0.0.0` in the default in flood settings?

Security. By listening on `127.0.0.1` by default, Flood is secure by default. It makes sure that Flood doesn’t get exposed to Internet by accident.

onedr0p:
Thanks I understand now. I've updated the PR

I'm curious if you think a entrypoint shell script is worth writing to copy into the container and use in the command part?

e.g.

```sh
....
COPY entrypoint.sh /entrypoint.sh
ENTRYPOINT ["/sbin/tini","--"]
CMD ["/entrypoint.sh"]
```

```sh
#!/bin/sh

exec "Do logic here to set default args and append user provided args"

```

jesec:
I would like to move all the flags to entrypoint

> Thanks I understand now. I've updated the PR
> 
> I'm curious if you think I'm the future a entrypoint shell script is worth writing to copy into the container and use in the command part?
> 
> e.g.
> 
> ```shell
> ....
> COPY entrypoint.sh /entrypoint.sh
> ENTRYPOINT ["/sbin/tini","--"]
> CMD ["/entrypoint.sh"]
> ```
> 
> ```shell
> #!/bin/sh
> 
> exec "Do logic here to set default args and append user provided args"
> ```

Not necessary.

onedr0p:
I've updated the PR to add them all.

Maybe I can throw in a feature request for flood to support environment variables? That would help make configuration easier than passing args.

jesec:
> I've updated the PR to add them all to the entrypoint
> 
> Maybe I can throw in a feature request for flood to support environment variables? That would help make configuration easier than passing args.

`FLOOD_OPTION_*`?

onedr0p:
Yes that would be awesome if it's not hard to add!

jesec:
> Yes that would be awesome if it's not hard to add!

Isn’t it already there?

onedr0p:
Sorry I didn't see if any docs on it. Perhaps I'm blind 😅

Why are we passing args to the entrypoint then? We can easily update the examples to show his people can configure this with env rather than args.

jesec:
> Sorry I didn't see any docs on it. Perhaps I'm blind 😅
> 
> Why are we passing args to the entrypoint then? We can easily update the examples to show how people can configure this with env rather than args.

Depends. `docker run -it jesec/flood --some-args` is more convenient than `docker run -it -e FLOOD_OPTION_SOME_ARGS=true jesec/flood`. Though, in a chart or other config file, env may be a better option.

onedr0p:
I think env is the way to go imho, it's makes the dockerfile most compatible across any kind of container deployment methods. Lots of projects out there do it this way and only use args when required.

Currently if I wanted to use env config in Kubernetes it wouldn't be compatible out of the box because the entrypoint is hard coded with those args.

jesec:
> I think env is the way to go imho, it's makes the dockerfile most compatible across any kind of container deployment methods. Lots of projects out there do it this way and only use args when required.
> 
> Currently if I wanted to use env config in Kubernetes it wouldn't be compatible out of the box because the entrypoint is hard coded with those args.

Those default arguments don't need to be overridden. You can use env for others.

onedr0p:
I get that, but why can't they be set as ENV instead of args in the dockerfile? It would make it much easier to work with and look much nicer too!

jesec:
> I get that, but why can't they be set as ENV instead of args in the dockerfile? It would make it much easier to work with and look much nicer too!

I am OK with that. Feel free to change Dockerfile to use ENV for options, then.

onedr0p:
Sounds good, I'll try to update the readme with examples too.

PR #262, opened 2021-04-01
feat: use env instead of args in docker images

onedr0p:
## Description

This sets up the defaults in the docker container to use config from env instead of args

## Related Issue

As discussed here https://github.com/jesec/flood/pull/261

## Types of changes

### DIFF
--- a/distribution/containers/Dockerfile.distroless
+++ b/distribution/containers/Dockerfile.distroless
@@ -6,16 +6,20 @@ WORKDIR /root
 
 # Install Flood and dependencies to /bin
 RUN mkdir -p /root/sysroot/bin
+
 COPY ./artifacts artifacts
-RUN \ 
-    apk --no-cache add tini \
-    if [[ `uname -m` == "aarch64" ]]; \
+RUN if [[ `uname -m` == "aarch64" ]]; \
     then mv artifacts/flood-linux-arm64 flood; \
     elif [[ `uname -m` == "x86_64" ]]; \
     then mv artifacts/flood-linux-x64 flood; \
     fi
 RUN mv flood /root/sysroot/bin/flood
+
 COPY --from=busybox /bin/busybox_DF /root/sysroot/bin/df
+
+RUN apk --no-cache add tini-static
+RUN cp /sbin/tini-static /root/sysroot/bin/tini
+
 RUN chmod 0555 /root/sysroot/bin/*
 
 # Create 1001:1001 user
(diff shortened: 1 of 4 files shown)


jesec:
I think we should keep the args way in the doc, while suggesting env as an alternative.

jesec:
And, can we use multiple ENV and move options near the entry point?

distribution/containers/Dockerfile.distroless
```diff
@@ -42,5 +49,15 @@ FROM flood as rtorrent-flood
...
+ENV \
+    HOME=/home/download \
+    FLOOD_OPTION_HOST="0.0.0.0" \
+    FLOOD_OPTION_PORT="3000" \
+    FLOOD_OPTION_RTORRENT="true"
+
 # Flood with managed rTorrent daemon
-ENTRYPOINT ["flood", "--host=0.0.0.0", "--rtorrent"]
+ENTRYPOINT ["/sbin/tini", "--"]
+CMD ["flood"]
```
jesec, line 63: we still probably want to use `flood` as the entrypoint.

onedr0p: Updated

onedr0p:
> And, can we use multiple ENV and move options near the entry point?

Do you really want separate docker layers for each ENV? Currently it is treated as 1 layer.

jesec:
> > And, can we use multiple ENV and move options near the entry point?
> 
> Do you really want separate docker layers for each ENV? Currently it is treated as 1 layer.

It actually makes more sense, as each ENV is not related to others. I am OK with zero-size layers, and that looks better in the Dockerfile.

jesec:
the wording looks a bit weird to me. I will probably prefer to keep the original one and add something like "Alternatively, you may use env ...".

onedr0p:
Now that env is default in the dockerfiles you still want to point people to using args over env?

jesec:
> Now that env is default in the dockerfiles you still want to point people to using args over env?

well. unless you think `docker ... -e FLOOD_OPTION_XXX=true ...` is more convenient than `docker ... --xxx`. 

It is OK to use env in chart or other file-based config. But for the command line, args is still the more user-friendly one.

onedr0p:
> unless you think `docker ... -e FLOOD_OPTION_XXX=true ...` is more convenient than `docker ... --xxx`.

For the complete docker beginner, no otherwise I do.

> It is OK to use env in chart or other file-based config. But for the command line, args is still the more user-friendly one.

But I don't think a lot of people are using `docker run` to run rtorrent, it's more more likely they are using docker-compose in which any method is fine but env is more widely used.

onedr0p:
But you are right we are describing this with `docker run` I'll update it

jesec:
Beginners exist and there are probably a lot. They probably don't know how to use Docker Compose. After all, one can have a poor man's service up with simple `nohup docker run ... &` in any shell.

distribution/containers/Dockerfile.release
```diff
@@ -21,8 +21,12 @@ RUN adduser -h /home/download -s /sbin/nologin --disabled-password download
 # Run as "download" user
 USER download
 
+ENV HOME=/home/download
```
jesec, line 24: For non-distroless (normal) images, it is unnecessary (and harmful) to override HOME inside the Dockerfile.

jesec: as there is one by default.

onedr0p: Updated

onedr0p:
Readme was updated

jesec:
`To view more available environment options see [here](https://github.com/jesec/flood/blob/master/config.ts).` 

Note that `FLOOD_OPTION_*` is simply env way to set a CLI argument. So `--help` is probably a more straightforward way to see help texts. Or run it with `FLOOD_OPTION_HELP=true` (but that's weird). And, there is already a section about `Configuration`: https://github.com/jesec/flood#configuration .

onedr0p:
Updated the link to point to https://github.com/jesec/flood#configuration

jesec:
I made some changes. Please check it out.

onedr0p:
Good catch on tini-static for distroless. Looks good

LGTM!

PR #316, opened 2021-05-25
feat: download .torrent file saves as human name intead of hash name

sabersalv:
## Description

For https://github.com/jesec/flood/discussions/241

I understand you like the hash name, then how about add a flag? 
Would love it if you can add a checkbox in the settings UI.

## Types of changes

### DIFF
--- a/server/routes/api/torrents.ts
+++ b/server/routes/api/torrents.ts
@@ -28,6 +28,7 @@ import type {
   StartTorrentsOptions,
   StopTorrentsOptions,
 } from '@shared/types/api/torrents';
+import type {TorrentListSummary} from '@shared/types/Torrent';
 
 import {
   addTorrentByFileSchema,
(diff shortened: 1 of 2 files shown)


jesec:
Hmm.. This is a bit bloated.

jesec:
In general, API should have well-defined behaviors. In this case, hash is well-defined and parsable but name is not. Additionally, it is possible to have torrents with duplicate names. 

I do agree that sometimes users may prefer a more human-friendly file name. I think we can implement that in single-file case with `download` anchor attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download . 

For the multi-file case, I don't think we can do much about it. It would be chaotic to have an API that returns a non-predicable archive.

sabersalv:
My usecase is: I want to bulk upload torrents to another site. So I'll download 10 torrents at a time, save them to the disk, then upload one by one to another site. Now I have 10 torrents with hash names, I can't link the torrent to a tv show or movie name, so I can't use them at all.

Hash is well-defined and parsable, that's true for scripting. But most of the case, I think users download the torrent for a human to use, not scripting. When you have more than one torrent with hashes, you won't be able to identify the torrent at all.

Since I only use a few torrents at a time, not thounds of thounds. duplicated name won't be a problem.

jesec:
> Tested with the `download` attribute. It didn't work as the doc said
> 
> > If the Content-Disposition header has different information from the download attribute, resulting behavior may differ:
> > If the header specifies a filename, it takes priority over a filename specified in the download attribute.

I see. Another way is to let the frontend download the torrent file to the memory first, and then spawn the download with designated name. Torrent files are quite small, so it should be OK to use this approach.

sabersalv:
closed until you can find a better way for this.

PR #318, opened 2021-05-25
fix: change the cookie sameSite to lax

sabersalv:
## Description

I'm building an web app that interact with Flood. And I want to implement this feature, the app list torrents with download torrent links (e.g. `http://flood/api/torrents/hash/metainfo`), when user click it, it follows the link to flood site and download the torrent file. According to [this article](https://web.dev/samesite-cookies-explained/), needs set the sameStite to `Lax`

With sameSite to `Strict`, the followed link will show `unauthorized` error.
With sameSite set to `Lax`, the followed link works

Combine with https://github.com/jesec/flood/pull/316, you can download the torrent and save it into a meaningfull name.

### DIFF
--- a/server/util/authUtil.ts
+++ b/server/util/authUtil.ts
@@ -10,7 +10,7 @@ const EXPIRATION_SECONDS = 60 * 60 * 24 * 7; // one week
 export const getCookieOptions = (): CookieOptions => ({
   expires: new Date(Date.now() + EXPIRATION_SECONDS * 1000),
   httpOnly: true,
-  sameSite: 'strict',
+  sameSite: 'lax',
 });
 
 export const getAuthToken = (username: string, iat?: number): string => {


jesec:
The strict sameSite policy is intentional: https://github.com/jesec/flood/commit/2984d58e2d0c04a5801ba0458f03c05fc1b8199a .

If your website can utilize such action, unfortunately, malicious websites can do that as well. 

I suggest you to explore alternatives. Note that if your web application performs authentication with Flood's backend, likely you would already have the necessary tokens, and you don't have to use the one from the user's session. Additionally, Electron allows you to exempt your app from certain security policies, and browser extensions are not bound by the policies.

sabersalv:
Lax is as secure as the Strict.

1. call fetch won't include the cookie. Same as Strict.
2. embeded an image won't include the cookie. Same as Strict.
3. When you click the link, it opens in a new tab and goes to flood site, the cookie is included. This behaivor is same as you enter the link in the brower address bar and goes to flood site. The app has no control of the behavior, there's no way for the app to access the cookie.

jesec:
> Lax is as secure as the Strict.
> 
> 1. call fetch won't include the cookie. Same as Strict.
> 2. embeded an image won't include the cookie. Same as Strict.
> 3. Only when you click the link, it opens in a new tab and goes to flood site,  the cookie is included. This behaivor is same as you enter the link in the brower address bar and goes to flood site. The app has no control of the behavior, there's no way for the app to access the cookie.

Not really. Otherwise, `Strict` would not exist at all. It is there for a reason. 

Some Flood API endpoints can be sensitive or state-changing. `Strict` prevents malicious websites from firing the top-level navigation to perform undesired actions with those endpoints. 

You should try to explore alternatives first. As I mentioned:

- Electron app -> tweak security parameters
- Browser extension -> generally not limited by such security policy
- App with backend -> send the request from backend instead

sabersalv:
Okay, I'll explore the alternatives. Mine is a web application, so point 3 and 4 are the solutions. I would avoid point 3 for the simplicity. Now left point 4, as it's a web application, it leads back to [this issue](https://github.com/jesec/flood/pull/228#issuecomment-792321678), web app can't get auth token by authenticate with the Flood API directly. Maybe it's time to reconsider addding an option to include the token in other places in /auth response?

jesec:
> Okay, I'll explore the alternatives. Mine is a web application, so point 3 and 4 are the solutions. I would avoid point 3 for the simplicity. Now left point 4, as it's a web application, it leads back to [this issue](https://github.com/jesec/flood/pull/228#issuecomment-792321678), web app can't get auth token by authenticate with the Flood API directly. Maybe it's time to reconsider addding an option to include the token in other places in /auth response?

Well. How would your app work if it can't authenticate with Flood? Which part of your app do the real communication with Flood API?

If the backend does that, you have to stick to the backend. You can't directly call the Flood API in any way in the frontend.

sabersalv:
Looks like the only way is through backend. closed.

I wish Flood can add advanced options (like Firefox about:config hidden page) for advanced users. Make API avaialble to the frontend can bring tons of possiblilities for developers to build amazing apps around the Flood, make the whole ecosystem bigger.

jesec:
The security policy is there to prevent untrusted websites from accessing Flood API. It does not hurt the normal use case in any way. 

In your use case, if the backend does the authentication, it would be inappropriate to assume that the browser's Flood session is the same as the backend's session, as they may be logged in to different users. Additionally, backend's access to Flood does not imply that the user's browser can access Flood (in the same way). For example, your app's backend may access Flood on `localhost`, but likely the user's browser can't if it accesses your app from another machine.

PR #327, opened 2021-05-30
server: partially migrate to fastify

jesec:
[expressjs/express](https://github.com/expressjs/express) is unmaintained.


trim21:
I can work on this

PR #375, opened 2021-07-15
fix: RSS feed rule edit

sabersalv:
## Description

1. Edit a RSS rule calls `removeItem`, then `addRule`
2. `addRule` keep pushing rule into `this.rules`

https://github.com/jesec/flood/blob/6aac6486efbb360a24b6b8370156928cf5f6b910/server/services/feedService.ts#L181 

3. But `removeItem` does not remove rule from `this.rules`

4. That caused the problem, edit a rule may not work. When you edit the rule multpile times, `this.rules` becomes larger and larger, old rules are never get removed.

In the future, can create a new `editRule` api for simplication.

## Types of changes

### DIFF
--- a/server/services/feedService.ts
+++ b/server/services/feedService.ts
@@ -1,5 +1,6 @@
 import path from 'path';
 import Datastore from 'nedb-promises';
+import {remove} from 'lodash';
 
 import type {FeedItem} from 'feedsub';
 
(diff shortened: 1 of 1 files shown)


jesec:
Avoid `lodash`. If you do need to use it, import that function only.

sabersalv:
@jesec Fixed. By the way, any particular you want to avoid lodash?

jesec:
> @jesec Fixed. By the way, any particular you want to avoid lodash?

It is heavy. Once you import it, the bundle size increases by 50-ish KBs. webpack is not able to tree-shake the unused methods out by default. You should import it with syntax like `import remove from 'lodash/remove'` instead of `import {remove} from 'lodash'`.

PR #383, opened 2021-07-31
feat: display selected count and size on context menu

sabersalv:
## Description

When you selected multiple items, righ click will give you information about the selected count and size. 
The info will not show up if you only selected 1 item.
You can enable/disable it in the settings.

A very useful feature when you want to know how many items you have selected.

## Screenshots

Screen Shot 2021-07-31 at 3 04 01 PM

## Types of changes


jesec:
hmm... It is rather unusual to see the count in context menu.

Would it be better to display that on action bar? It would make sense to utilize the empty space between the "start/stop" buttons and "collapse sidebar" button.

sabersalv:
Closed in favor of #385

PR #384, opened 2021-07-31
feat: add select all on context menu

sabersalv:
## Description

Provide a way to select all, deselect all for mobile device. Can enable/disable it in the settings.

Depends on #383

## Screenshots

Screen Shot 2021-07-31 at 5 49 41 PM
Screen Shot 2021-07-31 at 5 49 47 PM

## Types of changes


jesec:
I don't think it is a good idea to bloat the context menu. Is there a better way? 

 Maybe we can add a button to action bar instead.

sabersalv:
Closed in favor of #386

PR #385, opened 2021-08-02
feat: display selected count and size on action bar

sabersalv:
## Description

Will show up only if number of the selected items are larger than or equal to 2.

## Screenshots

Screen Shot 2021-08-02 at 11 59 47 PM
Screen Shot 2021-08-03 at 12 00 49 AM
Screen Shot 2021-08-03 at 12 01 00 AM
Screen Shot 2021-08-03 at 12 04 11 AM
Screen Shot 2021-08-03 at 12 04 34 AM
Screen Shot 2021-08-03 at 12 04 41 AM

## Types of changes

### DIFF
--- a/client/src/javascript/components/torrent-list/ActionBar.tsx
+++ b/client/src/javascript/components/torrent-list/ActionBar.tsx
@@ -1,7 +1,7 @@
 import classnames from 'classnames';
 import {FC} from 'react';
 import {observer} from 'mobx-react';
-import {useLingui} from '@lingui/react';
+import {useLingui, Trans} from '@lingui/react';
 
 import {Add, Menu, Remove, Start, Stop} from '@client/ui/icons';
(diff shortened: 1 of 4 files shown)


jesec:
Hmm.... That does not look good on smaller screens. We should probably hide it when screen width is smaller than a certain value. 

It also feels weird on expanded view. The size should scale to match other components on the bar.

sabersalv:
Fixed

Screen Shot 2021-08-06 at 12 10 21 AM

Hide the count for screens smaller than 375px. I was trying to add mulit select for mobile, found out it's too complicated. I added a [select all button](https://github.com/jesec/flood/pull/386) instead. Mobile can still filter the torrents by sidebar (for example, view count with a tracker + a tag)

sabersalv:
Closed until you can think of a better design for this.

PR #386, opened 2021-08-03
feat: add select all button on action bar

sabersalv:
## Screenshots

Screen Shot 2021-08-03 at 7 13 05 PM

## Types of changes

### DIFF
--- a/client/src/javascript/actions/UIActions.ts
+++ b/client/src/javascript/actions/UIActions.ts
@@ -37,6 +37,10 @@ const UIActions = {
     TorrentStore.selectAllTorrents();
   },
 
+  deselectAllTorrents: () => {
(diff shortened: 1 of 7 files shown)


jesec:
There are too many irrelevant changes...

On a second thought, I do feel that the action bar looks a bit too bloated with an additional button. I don't have a good idea about the alternative at the moment though.

sabersalv:
Rebased, I hide the button on desktop. Will only show on mobile devices, cause mobile lacks of select all.

sabersalv:
Closed until you can find a better design to fit this.

jesec:
Yeah. My primary concern is that the bloated action bar would make an existing issue even worse on mobile devices.

PR #417, opened 2021-10-03
CI: add "jesec/flood:latest" tag to rolling release

DopyConqueror:
## Description

Docker's default tag is "latest", so when not specifying a tag when pulling/pushing a tag from/to Docker Hub "latest" is used. The "latest" tag was used for "jesec/flood" previously but is not used anymore, leaving an outdated tag on Docker Hub. This outdated tag gets downloaded by default. This affects some of the docker-compose examples, too.

This PR adds the "latest" tag as a synonym for the "master" tag.

## Types of changes

### DIFF
--- a/.github/workflows/publish-rolling.yml
+++ b/.github/workflows/publish-rolling.yml
@@ -87,14 +87,16 @@ jobs:
         run: |
           echo ::set-output name=BUILD_VERSION::0.0.0-master.`git rev-parse --short HEAD`
 
-      - name: Publish flood:master to Docker Hub
+      - name: Publish flood to Docker Hub
         uses: docker/build-push-action@v2
         with:
           context: .
           file: ./distribution/containers/Dockerfile.release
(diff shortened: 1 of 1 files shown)


jesec:
No. It is intended. `latest` should point to a released build.

DopyConqueror:
Then `jesec/rtorrent-flood:latest` should point to a release build, too.

DopyConqueror:
There also needs to be a new release, because `latest` doesn't include the new Let's Encrypt root certificate, breaking a lot of HTTPS trackers.

jesec:
4.7 has been released.

DopyConqueror:
> 4.7 has been released.

Thank you!

Also I noticed that `jesec/rtorrent-flood` only tracks master so that's the only thing `latest` can point to. Since all confusion has been cleared up and this PR does not implement desired behavior I'm closing it.

PR #461, opened 2021-12-12
Force new component instance when editing download rules and feeds

### LINKED ISSUE #460: Input states do not update when changing currently edited download rule or feed
Type: Bug Report

- [x] Try to follow the update procedure described in the README and try again before opening this issue.

## Your Environment
- Version used:
  - master branch
- Environment name and version:
  - Node.js version `14`
  - npm version `6.14.15`
  - Web browser `Firefox 95`
- Operating system and version: Linux Debian Buster
- Torrent client and version: rtorrent 0.9.8 (jesec/rtorrent)

## Summary

Editing download rules/feeds does not update the state of the textboxes/checkboxes when switching between currently edited item.

## Expected Behavior

Pressing the edit icon should update all the input fields to the values of newly selected item

## Current Behavior

Input contents retain the values of the first item that was edited. Hitting save also updates the new item with previously opened item values.

## Possible Solution

Looking at the source code, I guess the problem lies in the defaultValue passed to the text boxes. When the item changes, these values are ignored on already rendered fields and thefor take no effect.

The submission happens in the parent component on the `<Form>` component where the item is updated. So the submission process takes the current values of inputs and then submits it towards the item ID that was selected last.

Changing the item prop should either rerender the fields or the fields should provide a two way binding on the values.

I'll see if I can get to submitting a PR

## Steps to Reproduce

1. Open the modal with feeds
2. Go to rules/feed list
3. Create two items with different values
4. Click the edit icon next to item A
5. Click the edito icon next to item B
6. Item A should retain the values from item B
7. Save the item B
8. Reload the page/close feeds, browse to the item B, it should have the same values are item A

Sharsie:
## Description

Due to text inputs retaining their state on component rerenders
editing the download rules and feeds forces previous values to be bound
to the newly edited rule/feed.
A new component instance is forced by keying the parent component
and thus the new defaultValue on inputs is taken into account

## Related Issue

fixes #460

## Types of changes

## Comments

This is honestly a lazy fix, I think there could be other related issues to this due to the way `<form>` onSubmit is used and defaultValues on inputs to initialize them. Similar problems could arise anywhere else in the application where list-editing is handled in the same way, so I'd recommend a different aproach, but don't want to make too much of a mess.

### DIFF
--- a/client/src/javascript/components/modals/feeds-modal/DownloadRulesTab.tsx
+++ b/client/src/javascript/components/modals/feeds-modal/DownloadRulesTab.tsx
@@ -198,6 +198,7 @@ const DownloadRulesTab: FC = () => {
       </FormRow>
       {isEditing ? (
         <DownloadRuleForm
+          key={currentRule?._id || 'initial'}
           rule={currentRule ?? initialRule}
           isPatternMatched={isPatternMatched}
           isSubmitting={isSubmitting}
--- a/client/src/javascript/components/modals/feeds-modal/FeedsTab.tsx
+++ b/client/src/javascript/components/modals/feeds-modal/FeedsTab.tsx
@@ -178,6 +178,7 @@ const FeedsTab: FC = () => {
           <FeedForm
             currentFeed={currentFeed}
             defaultFeed={defaultFeed}
+            key={currentFeed?._id || 'initial'}
             intervalMultipliers={INTERVAL_MULTIPLIERS}
             isSubmitting={isSubmitting}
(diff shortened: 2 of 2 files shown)


Sharsie:
Actually, following the comments above, I just realized another list editing is implemented in list of feeds, fixing in a sec

Sharsie:
Rebased on master, fixes both instances where I found the issues, let me know if you know of any other place or if you want to tackle this more thoroughly as stated above.

PR #518, opened 2022-02-20
Added filter by filesystem location + bonus tooltip on overflowed filter names

FinalDoom:
New filter section that allows viewing space usage/torrent count by directory in tree format.
Filter names overflow with ellipses, and in the case of root directories, this is common: so I added a title-based tooltip to filter elements where the text is overflowed. Only appears on overflowed elements, and on mouse event. Not sure if equivalent touch events would work the same way.

No issue submitted, just something I wanted to make.

image
image

### DIFF
--- a/server/services/taxonomyService.ts
+++ b/server/services/taxonomyService.ts
@@ -3,7 +3,7 @@ import jsonpatch, {Operation} from 'fast-json-patch';
 import BaseService from './BaseService';
 import torrentStatusMap from '../../shared/constants/torrentStatusMap';
 
-import type {Taxonomy} from '../../shared/types/Taxonomy';
+import type {Taxonomy, LocationTreeNode} from '../../shared/types/Taxonomy';
 import type {TorrentStatus} from '../../shared/constants/torrentStatusMap';
 import type {TorrentProperties, TorrentList} from '../../shared/types/Torrent';
 
@@ -13,6 +13,9 @@ interface TaxonomyServiceEvents {
 
 class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
   taxonomy: Taxonomy = {
+    locationCounts: { '': 0 },
+    locationSizes: {},
+    locationTree: [],
     statusCounts: {'': 0},
     tagCounts: {'': 0, untagged: 0},
     tagSizes: {},
@@ -61,6 +64,9 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
 
   handleProcessTorrentListStart = () => {
     this.lastTaxonomy = {
+      locationCounts: {...this.taxonomy.locationCounts},
+      locationSizes: {...this.taxonomy.locationCounts},
+      locationTree: [...this.taxonomy.locationTree],
       statusCounts: {...this.taxonomy.statusCounts},
       tagCounts: {...this.taxonomy.tagCounts},
       tagSizes: {...this.taxonomy.tagSizes},
@@ -72,6 +78,9 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
       this.taxonomy.statusCounts[status] = 0;
     });
 
+    this.taxonomy.locationCounts = {'': 0};
+    this.taxonomy.locationSizes = {};
+    this.taxonomy.locationTree = []
     this.taxonomy.statusCounts[''] = 0;
     this.taxonomy.tagCounts = {'': 0, untagged: 0};
     this.taxonomy.tagSizes = {};
@@ -79,9 +88,58 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
     this.taxonomy.trackerSizes = {};
   };
 
-  handleProcessTorrentListEnd = ({torrents}: {torrents: TorrentList}) => {
-    const {length} = Object.keys(torrents);
+  buildLocationTree = () => {
+    const locations = Object.keys(this.taxonomy.locationCounts);
+    const sortedLocations = locations.slice().sort((a, b) => {
+      if (a === '') {
+        return -1;
+      }
+      if (b === '') {
+        return 1;
+      }
 
+      // Order slashes before similar paths with different symbols, eg. /files/PC/ should come before /files/PC-98/ for treeing
+      return a.replace(/[^\\/\w]/g, "~").localeCompare(b.replace(/[^\\/\w]/g, "~"));
+    });
+
+    const separator = sortedLocations.length < 2 || sortedLocations[1].includes('/') ? '/' : '\\';
+    let previousLocation: LocationTreeNode;
+    this.taxonomy.locationTree = sortedLocations.reduce((tree, filter) => {
+      const directory = filter.split(separator).slice(-1)[0];
+      const parentPath = filter.substring(0, filter.lastIndexOf(separator + directory))
+      const location: LocationTreeNode = { directoryName: directory, fullPath: filter, children: [] };
+      while (previousLocation) { // Move up the tree to a matching parent
+        if (!previousLocation.parent || previousLocation.fullPath === parentPath) {
+          break;
+        }
+        previousLocation = previousLocation.parent;
+      }
+      if (previousLocation && previousLocation.fullPath === parentPath && parentPath !== '') { // Child
+        location.parent = previousLocation;
+        previousLocation.children.push(location);
+      } else if (previousLocation && previousLocation.parent && previousLocation.parent.fullPath === parentPath) { // Sibling
+        location.parent = previousLocation.parent;
+        previousLocation.parent.children.push(location);
+      } else { // Root
+        tree.push(location);
+      }
+      previousLocation = location;
+      return tree;
+    }, [] as LocationTreeNode[]);
+  }
+
+  cleanLocationTree = (location: LocationTreeNode) => {
+    location.parent = undefined;
+    location.children.forEach(this.cleanLocationTree);
+  }
+
+  handleProcessTorrentListEnd = ({ torrents }: { torrents: TorrentList }) => {
+    const { length } = Object.keys(torrents);
+
+    this.buildLocationTree();
+    this.taxonomy.locationTree.forEach(this.cleanLocationTree);
+
+    this.taxonomy.locationCounts[''] = length;
     this.taxonomy.statusCounts[''] = length;
     this.taxonomy.tagCounts[''] = length;
     this.taxonomy.trackerCounts[''] = length;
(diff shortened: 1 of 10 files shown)


mouzzampk2014:
This is brilliant, thanks for working on this.

jesec:
The general idea looks OK, but the operations could get prohibitively expensive when there are many entries.

FinalDoom:
Open to other algorithms, but I can attest it takes minimal amount of time more than master with ~6k torrents.

jesec:
> Open to other algorithms, but I can attest it takes minimal amount of time more than master with ~6k torrents. image And with 10^5 directories (actual insanity IMO); 10 directories each with 10 children, each with 10 children, and so on, 5 deep. image

Numbers look acceptable, but I would assume there is a non-trivial amount of associated CPU usage. I am reluctant to add a potentially expensive operation to the poll routine. We can probably employ more memoization.

> Would it be worth making a default-disabled setting to enable the feature?

Not a good idea when the server-side functions are involved. There could be API concerns. 

I am tied up with some other matters at hand, and I will take a deeper look next week.

boeto:
When will it be merge? Expect this awesome feature

trim21:
can you resolve conflicts？

FinalDoom:
I can, are you looking to actually merge it? (don't wanna play the fix/wait/fix/wait game)

trim21:
> I can, are you looking to actually merge it? (don't wanna play the fix/wait/fix/wait game)

yes

FinalDoom:
Should be good now, also changed the mouseover to a hook, should be nicer.

trim21:
give me some time to test this locally

trim21:
this doesn't look right, in my understand shouldn't this display like a tree? some elements are displayed at top level


image
image

trim21:
also, just a question, is it possible to build filter at client instead of server like tags?

FinalDoom:
> shouldn't this display like a tree or multiple trees?

I'll have to take a look at it again, as it's been a while; but iirc the trees are built by comparing common paths, removing prefixes, and traversing down. I think the paths come from those reported on the torrent metadata.. so is it possible those torrents are using relative paths or something starting with the hex? 

Which root path should those non-root elements be under?

> is it possible to build trees at client instead of server like tags?

Probably? I'm guessing I mimicked something that needed to be server-side, but again it's been a while. I'll have to take a look at what metadata ends up on the client.

trim21:
> so is it possible those torrents are using relative paths or something starting with the hex?
>
> Which root path should those non-root elements be under?

no, they are all abs path.

My torrents' download directories:

- `/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/{hex}`
- `/srv/dev-disk-by-uuid-3c509d6a-504e-4d71-a19d-5f81c62d9902/...`
- `/export/3t-1/downloads/...`

so all hex elements and elements starts by `dev-disk-by-uuid` are non-root elements

FinalDoom:
Very odd. I'll have to take another look and make some test cases when I get a chance.

trim21:
> Very odd. I'll have to take another look and make some test cases when I get a chance.

minimal reproduction: https://gist.github.com/trim21/429853a67f294ba8a0c162488a2822b9

trim21:
looks like I have mixed case directories and they cause this bug:

- `/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/U2-small/e9b3a6c75170e0c14701e7095b94412cfd4c7c7d`
- `/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/00ac7be5eb069dd8af992fca3a40b3755aaad846`

trim21:
```js
console.log(
  [
    '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/231d4f952201ba96628a9520740554c6ae6837db',
    '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/U2-small/232e75cf79187a07a7507f82275f1e8d3f7ddf73',
    '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/2364ebde4155eaf3267af1011b4dbc9cc75d99e1',
  ].sort((a, b) => {
    return a.localeCompare(b, undefined, {sensitivity: 'case'});
  }),
);
```
=>
```js
[
  '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/231d4f952201ba96628a9520740554c6ae6837db',
  '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/U2-small/232e75cf79187a07a7507f82275f1e8d3f7ddf73',
  '/srv/dev-disk-by-uuid-58b9efe4-428e-40d1-8d3f-97d8a5bac12a/u2-small/2364ebde4155eaf3267af1011b4dbc9cc75d99e1'
]
```

server/services/taxonomyService.ts
```diff
@@ -72,16 +78,72 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
...
+    const sortedLocations = locations.slice().sort((a, b) => {
+      if (a === '') {
+        return -1;
+      }
+      if (b === '') {
+        return 1;
+      }
+
+      // Order slashes before similar paths with different symbols, eg. /files/PC/ should come before /files/PC-98/ for treeing
+      return a.replace(/[^\\/\w]/g, '~').localeCompare(b.replace(/[^\\/\w]/g, '~'));
```
trim21, line 102:
```suggestion
      const aa = a.replace(/[^\\/\w]/g, '~');
      const bb = b.replace(/[^\\/\w]/g, '~');
      if (aa < bb) {
        return -1;
      }
      return 1;
```

trim21: should fix this bug

FinalDoom: Ah ha, that would do it. Are you running a case-insensitive OS in some manner that's programmatically detectable? Windows is obviously.. Looks like mac is too? Linux might be in a rare case of custom ext4. 

I'm thinking it might be easiest to just add `i` to the regex flags if on a case-insensitive OS... or it looks like even better is passing options to localeCompare.

```suggestion
      const insensitiveOS = ['Darwin', 'Windows_NT'].includes(os.type());
      const compareOptions = {ignorePunctuation: true};
      if (insensitiveOS) {
        compareOptions['sensitivity'] = 'accent';
      }
      // I'm not sure what case I had for replacing everything but word characters and slashes with '~' -- that also drops non-latin characters, which isn't great.
      return a.localeCompare(b, undefined, compareOptions);
```

FinalDoom: Obviously a handful of test cases for this is in order.

trim21: won't work

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#sensitivity

FinalDoom: Why won't it work? Mac and Windows are insensitive, so for insensitive OS, it uses 'accent', which is case insensitive.

trim21: With 'accent', this array is still not sorted as expected.

trim21: In my understanding the root case is array being sorted case-insensitive, so we don't need to distinguish os, just need to sort them case-sensitive

trim21: and also, I'm running flood on a case-sensitive OS, so...

trim21: I just test this locally, it doesn't work.

server/services/taxonomyService.ts
```diff
@@ -97,13 +167,48 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
...
     this.incrementTrackerSizes(torrentProperties.trackerURIs, torrentProperties.sizeBytes);
   };
 
+  incrementLocationCountsAndSizes(
+    directory: TorrentProperties['directory'],
+    sizeBytes: TorrentProperties['sizeBytes'],
+  ) {
+    const separator = directory.includes('/') ? '/' : '\\';
+    const parts = directory.startsWith(separator) ? directory.split(separator).slice(1) : directory.split(separator);
+    let heirarchy = '';
```
trim21, line 184:
```suggestion
    let hierarchy = '';
```

server/services/taxonomyService.ts
```diff
@@ -61,6 +64,9 @@ class TaxonomyService extends BaseService<TaxonomyServiceEvents> {
 
   handleProcessTorrentListStart = () => {
     this.lastTaxonomy = {
+      locationCounts: {...this.taxonomy.locationCounts},
+      locationSizes: {...this.taxonomy.locationCounts},
+      locationTree: [...this.taxonomy.locationTree],
```
trim21, line 69:
```suggestion
      locationTree: _.cloneDeep(this.taxonomy.locationTree),
```

trim21:
lgtm, only one minor problem

FinalDoom:
I have a reduced solution that can probably be moved client-side; but case-insensitivity causes problems when filtering torrents. I think the solution is for the tree to be client-side, and the server can tell the client if the OS it's running on is case-insensitive or not, allowing insensitive tree and filtering

The current solution means client has no clue on insensitivity, so filtering on non-matching lowercased names omits torrents incorrectly. 

Another option would be to always treat everything as case sensitive, and you'd end up with duplicate paths where casing doesn't match.

But it's late, so I'll give it a shot tomorrow.

trim21:
> I have a reduced solution that can probably be moved client-side; but case-insensitivity causes problems when filtering torrents. I think the solution is for the tree to be client-side, and the server can tell the client if the OS it's running on is case-insensitive or not, allowing insensitive tree and filtering
> 
> The current solution means client has no clue on insensitivity, so filtering on non-matching lowercased names omits torrents incorrectly.
> 
> Another option would be to always treat everything as case sensitive, and you'd end up with duplicate paths where casing doesn't match.
> 
> But it's late, so I'll give it a shot tomorrow.

no need to move it to client side， current code looks fine

trim21:
handling case-insensitivity may introduce too much extra complexity， I think current code works fine

server/services/taxonomyService.ts
trim21: please resolve this one

FinalDoom:
I looked a bit at moving the treeing to client-side:

I'm guessing there's some reason for building the tag list and other sidebar filters server-side, whether it's shared efficiency for multiple clients, the taxonomy diff data savings, or something else. I reduced the complexity of computing the tree and just stored the counts/data size in-tree, so it should be a little faster/simpler overall.

There may be an argument for moving filter value computation to the client, but I think matching the existing sidebar filter logic makes more sense for now.

server/services/taxonomyService.ts
FinalDoom: done

trim21:
please do not use force pushing

FinalDoom:
I don't see what's failing in the tests. It names the new test file but there's no failing test or other message. It passes locally. Github doesn't have a way to rerun workflows?

FinalDoom:
> please do not force push

It's in my repo and rebasing creates better changeset history, why not?

trim21:
> > please do not force push
> 
> It's in my repo and rebasing creates better changeset history, why not?

Force-push causes inconvenient in code reviewing

server/util/fileUtil.ts
```diff
@@ -37,10 +37,7 @@ export const isAllowedPath = (resolvedPath: string) => {
   }
 
   return config.allowedPaths.some((allowedPath) => {
-    if (realPath?.startsWith(allowedPath)) {
-      return true;
-    }
-    return false;
+    return !!realPath?.startsWith(allowedPath);
```
trim21, line 40: please do not touch this file

trim21 (changes requested):

jiangjiangflies:
Can this feature be implemented? Really need it!

FinalDoom:
@trim21 you're welcome to edit it and resolve conflicts on my behalf. There is no logic change in the contested file, and the conflicts are in import ordering. All auto format results. 

Otherwise, I'll look at resolving the pr someday when I feel like it.

PR #519, opened 2022-02-20
Make sidebar filter sections collapseable

FinalDoom:
Makes sidebar filter sections collapseable by clicking the title. Since filter lists can get quite long, scrolling down to the tracker list or whichever can be tedious. Simple chevron/carat indicator hides filter section children.

image

### DIFF
--- a/client/src/sass/components/_sidebar-filter.scss
+++ b/client/src/sass/components/_sidebar-filter.scss
@@ -42,7 +42,7 @@
       }
     }
 
-    &--expando {
+    &--heading {
       cursor: default;
       font-size: 0.8em;
       font-weight: 500;
--- a/client/src/sass/components/_sidebar-filter.scss
+++ b/client/src/sass/components/_sidebar-filter.scss
@@ -42,7 +42,7 @@
       }
     }
 
-    &--heading {
+    &--expando {
       cursor: default;
       font-size: 0.8em;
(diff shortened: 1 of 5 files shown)


mouzzampk2014:
If its possible instead I collapsing all, UI wise might be idea thay after 10 or 15 trackers there will be a `more` at the end.

FinalDoom:
Probably doable simply enough @mouzzampk2014 . I'll keep it in mind when I feel like working on stuff again.

client/src/sass/components/_sidebar-filter.scss
```diff
@@ -42,7 +42,7 @@
       }
     }
 
-    &--heading {
+    &--expando {
```
jesec, line 45: Here, `sidebar-filter__item--heading` is basically removed.

If this is expected, the class should be removed from the `className` of components.

FinalDoom:
Good call, missed a bit on that cleanup. Restored --heading for consistency with other styles, and it was still used on disk usage.

client/src/sass/components/_sidebar-filter.scss
FinalDoom: Good call, missed a bit on that cleanup. Restored --heading for consistency with other styles, and it was still used on disk usage.

PR #523, opened 2022-03-09
Fix regression introduced in PR #519, fixes #522

### LINKED ISSUE #522: Issue on client side when no tags or trackers are defined on master
## Your Environment

- Version used:
  - Master branch
  - Commit ID (development release) : e368cac4fd0af4c4aeedad6bc609d66b97e60eb7
- Environment name and version:
  - Node.js version  16.14.0
  - npm version 8.3.1
  - Web browser Chrome
- Operating system and version: any
- Torrent client and version: rtorrent

## Summary

With the introduction of the features from https://github.com/jesec/flood/pull/519 , I have noticed that the overview page no longer works locally with no torrents. I have narrowed down the issue to the changes related to collapsible elements, and more precisely to the `useState` hooks that are declared after the components `Tag filters` and `Tracker filters` have returned when empty. 
This is a break of the rules of hooks that state that the list of hooks should always be the same and in the same order with every render.

## Expected Behavior

Page should be displayed properly

## Current Behavior

Page remains with a blue background

Capture d’écran 2022-03-09 à 23 40 04

## Possible Solution

#523 

## Steps to Reproduce

1. Launch the latest version of master on any environment
2. Start a fresh torrent session with no torrents
3. Log in with any working user

alexandresoro:
Fixes regression introduced in #519 that prevented to reach the overview page when no torrent was present.

The changes are simply to make sur that the expand status hooks are defined before the related components can be returned.

## Related Issue

#519 

## Types of changes

Fix is for a change that is not yest part of a release

### DIFF
--- a/client/src/javascript/components/sidebar/TagFilters.tsx
+++ b/client/src/javascript/components/sidebar/TagFilters.tsx
@@ -8,6 +8,7 @@ import TorrentFilterStore from '../../stores/TorrentFilterStore';
 
 const TagFilters: FC = observer(() => {
   const {i18n} = useLingui();
+  const [expanded, setExpanded] = useState<boolean>(true);
 
   const tags = Object.keys(TorrentFilterStore.taxonomy.tagCounts);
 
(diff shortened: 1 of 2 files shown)


jesec:
Thank you so much!

Yeah, this is a serious issue. It would have been caught by the frontend tests.

But unfortunately, they are disabled, pending the CSS refactoring. Fortunately, now I have more time available to restart working on that.

PR #533, opened 2022-03-22
Directly add torrents when paste valid magnet urls

fabricionaweb:
## Description

When you paste a valid magnet URL it will add the torrents directly.

It will capture all valid magnet urls from the pasted content (I'm using [this regex](https://github.com/tiaanduplessis/magnet-link-regex/blob/master/index.js)), and add then all.

## Types of changes


jesec:
It would be more useful to pop up the add window (with the magnet link pre-filled) instead.

fabricionaweb:
First I did it, then I changed :D

PR #539, opened 2022-04-02
Fix torrent download extension override

KaseiFR:
## Description

When downloading a single mkv file, Flood is changing its content-type to webm, and it makes Firefox change the file extension to .webm, which is annoying.
This PR fixes that by making the content-type override opt-in and only enabled in the file preview links.

I'd also like to add direct download links to the torrent files modal, next to the clipboard icon (or the reverse, make the file link a download and add a preview button). This would be useful when a torrent contains a few files that can be downloaded manually instead of going through the tar export. What do you think ?

## Types of changes


jesec:
This is another Firefox bug, which shall be reported to them. The extension declared by `content-disposition` should be prioritized over the MIME type. I will do it.

Meanwhile, we can use the User Agent condition for this.

PR #541, opened 2022-04-07
Added comment to torrent details

### LINKED ISSUE #435: Torrent Comments Missing from Torrent Details in Flood GUI
## Your Environment

`flood-linux-x64 4.7.0` with `rtorrent 0.9.8`

## Summary

I noticed that Flood doesn't display the Torrent Comment from the actual .torrent file anywhere in the interface.  As an easy example where this can be important, many of the top trackers include the URL to the individual torrent page, which can very useful, however they're not visible in Flood.  As an example here are the same torrents in ruTorrent and Flood:

rutorrent

flood

Also of note is that in ruTorrent the URL is clickable allowing you to return to the tracker torrent page in one click.

## Expected Behavior

Flood should show the Torrent Comment, and if a URL, make it clickable.

## Current Behavior

The Torrent Comment is not displayed anywhere in the interface.

## Steps to Reproduce

1. Add a .torrent file which contains a comment.
2. Open the details in Flood and see that there is no comment listed.

Thanks.

FinalDoom:
Mostly simple as it's supported by the various clients, except in the case of rtorrent.

For rtorrent, tags are stored in custom1, consistent with other clients. For that reason, comment is being stored in custom2,
which is also consistent with other clients. In particular, rutorrent uses a prefix on the comment for some reason, which is duplicated in this change to preserve cross-compatibility.

Also turns http: and https: text in comments into links, also consistent with other clients (and a main utility of having comments in the first place).

Fixes #435 

### DIFF
--- a/client/src/javascript/components/general/LinkedText.tsx
+++ b/client/src/javascript/components/general/LinkedText.tsx
@@ -1,0 +1,38 @@
+import {FC} from 'react';
+
+interface LinkedTextProps {
+  text: string;
+  className?: string;
+}
+
+function isValidHttpUrl(s: string) {
+  var url;
+
+  try {
+    url = new URL(s);
+  } catch (_) {
+    return false;
+  }
+
+  return url.protocol === 'http:' || url.protocol === 'https:';
+}
+
+const LinkedText: FC<LinkedTextProps> = ({text, className}: LinkedTextProps) => {
+  const nodes = text.split(/(?<=\s)(?!\s)(?:\b|\B)/).map((s) =>
+    isValidHttpUrl(s.trimEnd()) ? (
+      <a href={s.trimEnd()} target="_blank">
+        {s}
+      </a>
+    ) : (
+      s
+    ),
+  );
+
+  return <span className={className}>{nodes}</span>;
+};
+
+LinkedText.defaultProps = {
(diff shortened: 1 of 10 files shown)


jesec:
Where does the comment come from?

If it is set by the user, Flood users should be able to set it, as well. 

If it comes from the .torrent (metadata) file, I don't think rTorrent does the parsing, which means ruTorrent parses it. As such, Flood should parse it on add, as well. Better, add the command to rTorrent itself.

FinalDoom:
Precisely the latter. It comes from the torrent file. The non-rtorrent clients parse it themselves and report it themselves. While rTorrent does not. I mimicked the processing executed by ruTorrent for the rTorrent portion, to maintain consistency, and the comment is set in rTorrent by Flood.

Most clients/UIs allow setting the comment, as well, however I thought that would be too much for this objective. They also allow setting other parameters, such as torrent name (and data folder) which would use common logic/UI elements. So I may work on that in the future. Though it's a far less common use case, at least in my experience.

As for adding to rTorrent, are you referring to your repo of it? It looks like the rakshasha repo recieves little to no attention. I could take a look at adding it to your version, though C is my weakest language.

jesec (approved):

PR #593, opened 2022-09-05
i18n: small updates to the portuguese translation

saadbruno:
## Description

Quick updates to the portuguese translation. We don't use "BR" for gigabytes. This seems to have come from automatic translation.

## Types of changes


jesec:
Thanks for the PR.

However, translations are handled by Crowdin, and manual modifications would result in conflicts. Please use Crowdin to submit translations. 

https://crowdin.com/project/flood

PR #598, opened 2022-09-18
Allow torrent download directory to take more width

gillesdemey:
## Description

When using the "expanded view" and enabling the download location it only takes allows 100px to view the directory.

This change allows the directory detail to take up however much it needs – though perhaps it makes sense to add a sensible maximum width here, thoughts?

## Related Issue

N/A

## Screenshots

**before**
Screenshot 2022-09-18 at 16 11 58

**after**
Screenshot 2022-09-18 at 16 12 08

## Types of changes


trim21:
https://github.com/jesec/flood/pull/691

PR #610, opened 2022-11-08
Added a way to sort torrent peers columns

noefleury:
## Description

I've found sad that we didn't have the ability to sort the "Torrent Peers" table.

More precisely, currently it seems to be sorted automatically on the the backend side by IPs.

Personally, I'm always curious to see which peers are giving me the best download speeds and to where I'm uploading the faster.

## Screenshots

screenshot

## Types of changes

 So I've implemented a logic to sort columns on a click and I've followed the project logic to don't diverge too much from the main idea.

It is possible to sort by any columns and in any direction.

Note that the sort can handle IPv4/v6 too.

---

- [X] New feature (non-breaking change which adds functionality - semver MINOR)


trim21:
stalled, I'm going to replace this with https://github.com/jesec/flood/pull/1353

PR #622, opened 2022-12-21
Fix spelling in swedish translation

rasmusbe:
## Types of changes


rasmusbe:
Noticed that you are using crowdin, fixing it there

Issue #648, opened 2023-05-24
memory leak


trim21:
Type: Bug Report

- [X] Try to follow the update procedure described in the README and try again before opening this issue.

## Your Environment

- Version used: 7aec1e2
- Environment name and version:
  - Node.js version `node --version`: 18.16.0
  - npm version `npm --version`
  - Web browser `name and version`
- Operating system and version: windows/linux
- Torrent client and version: rtorrent

## Summary

memory leak in event source handler

## Steps to Reproduce

request dis-connect `"/activity-stream"` many time (about 1000)
you will find memory leaked, and web UI stop working

```python
import httpx
import sseclient
import tqdm


def stream():
    with httpx.stream("GET", "http://127.0.0.1:3000/api/activity-stream") as s:
        yield from s.iter_bytes()


def main():
    events = {
        "TORRENT_LIST_DIFF_CHANGE",
        "CLIENT_CONNECTIVITY_STATUS_CHANGE",
        "DISK_USAGE_CHANGE",
        "TAXONOMY_FULL_UPDATE",
        "TAXONOMY_DIFF_CHANGE",
        "TRANSFER_HISTORY_FULL_UPDATE",
        "TRANSFER_SUMMARY_FULL_UPDATE",
        "NOTIFICATION_COUNT_CHANGE",
        "TORRENT_LIST_FULL_UPDATE",
    }
    client = sseclient.SSEClient(stream())
    for _event in client.events():
        events.discard(_event.event)
        if events:
            return


if __name__ == "__main__":
    with tqdm.tqdm() as bar:
        while True:
            main()
            bar.update()
```

trim21:
very likely this is caused by rottrent request queue. there maybe a bug. I replace it with p-queue, and everythink looks fine.

ffactory-ofcl:
I'm on master and have memory leak problems. When running the flood development server, it quickly starts consuming upwards of 16 GB of memory. After some time, it either crashes or my Mac prompts me to close other applications, as I'm out of memory. 
This is the graph after closing flood (node):
<img width="194" alt="image" src="https://github.com/jesec/flood/assets/33554869/7f00c6e0-ec4a-40bb-85c6-4f3aa3018f5b">

trim21:
[...]

PR #650, opened 2023-06-11
fix rtorrent request queue memory leak

trim21:
## Description

use `p-queue` to queue requests

`p-queue` is a esm only package, so ts-node is replaced with [tsx](https://github.com/esbuild-kit/tsx)

`esbuild-jest` is added to transform esm only packages.

## Related Issue
close https://github.com/jesec/flood/issues/648

## Types of changes


trim21:
emmmmm, looks like jest doesn't work with esm only package

trim21:
@jesec

jesec:
This is great! Thank you, @trim21 . One thing I am very happy to see is that this simplified the logics a lot. I do have some wild concerns about introducing esbuild world of tools, but we can handle that later.

trim21:
> This is great! Thank you, @trim21 . One thing I am very happy to see is that this simplified the logics a lot. I do have some wild concerns about introducing esbuild world of tools, but we can handle that later.

what concerns do you have? for example, `tsx` doesn't perform type checking?

PR #653, opened 2023-06-15
Add tag (label) support for Deluge and adjust UI to accommodate

murtaza64:
## Description

Add tag support for Deluge (using the built in Labels plugin). Deluge only supports a single Label per torrent, so I introduce a new Tag selection UI preference "Single Strict" which allows you to select at most one tag when assigning tags.

To enforce this behavior for Deluge, the frontend checks the new readonly `clientSettings.tagSupport` and disables changing the tag UI mode if label support is limited to one-label-per-torrent.

Also fixes Deluge torrents showing 0B uploaded/0B downloaded.

TODO

- ~~skip RPCs if Labels plugin is inactive/unavailable~~ done

## Related Issue

#607 

## Types of changes

### DIFF
--- a/client/src/javascript/components/general/form-elements/TagSelect.tsx
+++ b/client/src/javascript/components/general/form-elements/TagSelect.tsx
@@ -100,7 +100,7 @@ const TagSelect: FC<TagSelectProps> = ({defaultValue, placeholder, id, label, on
             <ContextMenu
               isIn={isOpen}
               onClick={(event) => {
-                if (SettingStore.floodSettings.UITagSelectorMode !== 'single') {
+                if (SettingStore.floodSettings.UITagSelectorMode === 'multi') {
                   event.nativeEvent.stopImmediatePropagation();
                 }
               }}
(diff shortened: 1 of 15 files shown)


murtaza64:
Does anyone know a better way to communicate the tag support status (single tag or multiple tags) to the frontend than the method I used of creating a dummy client setting? The frontend seems to be completely client agnostic but it would be nice to pass some contextual information for per-client functionality.

Alternatively being able to know the torrent client in the frontend could be a solution, but would encourage client-specific behavior which may be something @jesec doesn't want.

murtaza64:
I haven't been able to test this with other clients to make sure nothing breaks--can someone check that tags work as before for non-Deluge?

FinalDoom:
I haven't used Deluge, but what's the single/multi distinction about? Iirc several torrent managers only have a single "label", and it's up to UI to do something with it or not. Why not just store a single label in deluge that's just the list passed from the client at set time, comma joined; and split it back out to send to the client as an array?

murtaza64:
PR is pretty stale and I don't use flood much anymore, so closing for now

PR #656, opened 2023-07-26
perf: mediainfo router should use async/await to get file info

trim21:
## Description

current mediainfo will block nodejs process, use async/await to avoid blocking

now we are targeting to node18, so also upgrade `"@types/node"` to `^18`

## Types of changes


server/services/qBittorrent/clientRequestManager.ts
```diff
@@ -483,7 +483,7 @@ class ClientRequestManager {
         new URLSearchParams({
           hashes: hashes.join('|').toLowerCase(),
           tags: tags?.join(','),
-        }),
+        } as Record<string, string>),
```
trim21, line 486: this and changes in `server/services/rTorrent/util/scgiUtil.ts` are caused by `@types/node`

jesec:
Looks good in general. Thanks for the PR.

Some nits:

- Don't bump the versionof Node.js type yet. We still want to maintain compatibility as much as possible. If we do want to bump, make clear what features require it.
- I think we can use `fs.promises.` instead of having a `fsp`

trim21:
> Looks good in general. Thanks for the PR.
> 
> Some nits:
> 
> * Don't bump the versionof Node.js type yet. We still want to maintain compatibility as much as possible. If we do want to bump, make clear what features require it.
> * I think we can use `fs.promises.` instead of having a `fsp`

OK， I revert version of `@types/node` back to 12

server/util/async.ts
```diff
@@ -0,0 +1,8 @@
+export async function asyncFilter<T>(
+  arr: Array<T>,
+  predicate: (item: T, index: number) => Promise<boolean>,
+): Promise<Array<T>> {
+  const results = await Promise.all(arr.map(predicate));
```
trim21, line 5: should use for loop and await to avoid too many promises maybe?

jesec: Yeah. That's a good idea.

trim21: then it's already done...

PR #657, opened 2023-07-27
feat: use relative path in mediainfo info

trim21:
## Description

make `mediainfo` generate info with relative path starts from torrent directory instead of absolute path, avoid leaking unexpected info.

previous:

```text
General
Complete name                            : /abs/path/to/filename
...
```

after:
```text
General
Complete name                            : filename
...
```

## Types of changes


server/routes/api/torrents.ts
```diff
@@ -905,8 +905,8 @@ router.get<{hash: string}>(
 
       const mediainfoProcess = childProcess.execFile(
         'mediainfo',
-        torrentContentPaths,
-        {maxBuffer: 1024 * 2000, timeout: 1000 * 10},
+        torrentContentPaths.map((x) => path.relative(torrentDirectory, x)),
+        {maxBuffer: 1024 * 2000, timeout: 1000 * 30, cwd: torrentDirectory},
```
trim21, line 909:
```suggestion
        {maxBuffer: 1024 * 2000, timeout: 1000 * 10, cwd: torrentDirectory},
```

PR #678, opened 2023-10-11
Add scripts for rss auto download

zoriya:
## Description

Allow users to define a script to decide if a rss item should be downloaded or not. If the scripts return the magic code of 80 (and regex match if it is specified), the item will be downloaded.

This PR also specifies a user agent of 'flood' for http requests used to retrieve the rss since some servers sent 403 when no user agent was specified.

## Types of changes

### DIFF
--- a/server/util/feedUtil.ts
+++ b/server/util/feedUtil.ts
@@ -1,3 +1,5 @@
+import {spawn} from 'node:child_process';
+
 import type {FeedItem} from 'feedsub';
 
 import type {AddTorrentByURLOptions} from '../../shared/schema/api/torrents';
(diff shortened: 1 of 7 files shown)


trim21:
there are several features mixed in this PR, would you mind split them?

zoriya:
You want one PR for the script & one for the header?

trim21:
there are also nix file and editor config file

trim21:
1. you can set miniget options with feedsub directly, there is no need to import miniget.
2. I don't know nix, I can just merge nix file, if you submit it in a seprated PR.
3. empty editorconfig doesn't make sense, please add some meanful content, if you want to submit it.

zoriya:
I removed the nix & editorconfig files, there was no need to push them mb.

I created a PR for the user agent https://github.com/jesec/flood/pull/779

server/util/feedUtil.ts
```diff
@@ -54,17 +55,33 @@ export const getTorrentUrlsFromFeedItem = (feedItem: FeedItem): Array<string> =>
...
+  });
+}
+
 export const getFeedItemsMatchingRules = (
   feedItems: Array<FeedItem>,
   rules: Array<Rule>,
 ): Array<PendingDownloadItems> => {
   return feedItems.reduce((matchedItems: Array<PendingDownloadItems>, feedItem) => {
-    rules.forEach((rule) => {
+    rules.forEach(async (rule) => {
```
trim21, line 78: this won't work

zoriya: I dont know how I never encountered a bug w/ this. I updated the function to be async

trim21:
can you add some example for potential users of this feature？

zoriya:
Sure, where do you want me to place this example?

I'm using the following script to download every item in the rss that has the same name as one of the existing folders:

```bash
#!/usr/bin/env bash
set -e

# Retrive name from rss title ($1) & remove special chars
name=$(guessit "$1" -P "title" | tr -d "[:punct:]")

# Check if we should download the item (if not grep will return 1)
ls /mnt/kyoo/shows/ | tr -d "[:punct:]" | grep -qix "$name"

echo "Downloading $1"
exit 80
```

PR #698, opened 2023-12-09
enable source map

trim21:
## Types of changes

### DIFF
--- a/.github/workflows/publish-rolling.yml
+++ b/.github/workflows/publish-rolling.yml
@@ -206,7 +206,7 @@ jobs:
 
       - name: Build executables
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/publish-rolling.yml
```diff
@@ -206,7 +206,7 @@ jobs:
 
       - name: Build executables
         run: |
-          pkg . --out-path dist-pkg
+          pkg . --out-path dist-pkg --public
```
trim21, line 209:
```suggestion
          pkg . --out-path dist-pkg --no-bytecode --public --options=enable-source-maps
```

PR #702, opened 2023-12-10
fix(client): react render logic

hmerritt:
# fix(client): react render logic

A couple react related fixes. One fixes a console error, the other an unwanted rendered result.

## Description

- Fix: `key={}` must be set on the rendered component (cannot be passed down and set within the component)
- Fix: conditional rendering via `&&` must be a conditional statement (not relying on value truthy/falsy) otherwise a `0` value will be rendered
  - Using `!=` to fix instead of `!==` because it catches both `null` and `undefined` values

## Related Issue

N/A

## Screenshots

Screenshot 2023-12-09 234933

Before / After
Screenshot 2023-12-09 235115
Screenshot 2023-12-09 235057

## Types of changes

### DIFF
--- a/client/src/javascript/components/sidebar/NotificationsButton.tsx
+++ b/client/src/javascript/components/sidebar/NotificationsButton.tsx
@@ -63,15 +63,14 @@ const NotificationTopToolbar: FC<NotificationTopToolbarProps> = ({
 };
 
 interface NotificationItemProps {
-  index: number;
   notification: Notification;
 }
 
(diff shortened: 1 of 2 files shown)


trim21:
how can I re-produce first image (before)?

hmerritt:
> how can I re-produce first image (before)?

If you add a torrent with a magnet link, the file-size is unknown before the metadata is downloaded (which usually happens quite quickly) - I had a torrent with basically 0 seeders so it never downloaded the metadata - and the filesize logic in the UI rendered `0`.

PR #705, opened 2023-12-11
fix(server): `@fastify/express` doesn't work with http2

trim21:
## Related Issue

close https://github.com/jesec/flood/issues/696

## Types of changes

### DIFF
--- a/server/bin/web-server.ts
+++ b/server/bin/web-server.ts
@@ -24,7 +24,6 @@ const startWebServer = async () => {
     instance = fastify({
       bodyLimit: 100 * 1024 * 1024,
       trustProxy: 'loopback',
-      http2: false,
       https: {
         allowHTTP1: true,
         key: fs.readFileSync(config.sslKey),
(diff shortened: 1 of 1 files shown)


server/bin/web-server.ts
```diff
@@ -24,7 +24,7 @@ const startWebServer = async () => {
     instance = fastify({
       bodyLimit: 100 * 1024 * 1024,
       trustProxy: 'loopback',
-      http2: true,
+      http2: false,
```
trim21, line 27:
```suggestion
```

server/bin/web-server.ts
```diff
@@ -24,7 +24,6 @@ const startWebServer = async () => {
     instance = fastify({
       bodyLimit: 100 * 1024 * 1024,
       trustProxy: 'loopback',
-      http2: true,
       https: {
         allowHTTP1: true,
```
trim21, line 28:
```suggestion
```

PR #717, opened 2024-01-03
Add config file flag

rasmus-kirk:
Adds a config file flag to load in config parameters from a JSON file.

## Description

Lets users load in command line configuration files as a JSON file using the `-c` or `--config` flags. Options compose, so options set by the command line and the config will combine. If the same option is set by both, the command line value takes precedence. Provided file _must_ be valid JSON and have the `.json` file extension. Tested by running with the `--config` flag set to a file containing:

```json
{
  "port": 3001
}
```

My use-case is that I'm using nix to configure Flood and providing the secret values for authentication as command line arguments would leak them to the nix store which means non-root users would be able to see the secrets.

### DIFF
--- a/config.ts
+++ b/config.ts
@@ -194,6 +194,8 @@ const {argv: argvObj} = yargs(process.argv.slice(2))
     hidden: true,
     type: 'boolean',
   })
+  .config()
+  .alias('c', 'config')
   .version(version)
   .alias('v', 'version')
   .help();


trim21:
thank s for your contribution

trim21:
is it possible to support yaml as well？

rasmus-kirk:
> is it possible to support yaml as well？

[Not really]([config](https://yargs.js.org/docs/#api-reference-configobject)), we can give it a custom parse function, but even _if_ that's doable, the file will still have to have the `.json` file extension.

trim21:
I saw config.txt in this PR

https://github.com/yargs/yargs/pull/321/files

rasmus-kirk:
I must be wrong then, well, in that case it should be possible given that we import a yaml parser. If you think yaml support is important I can try adding it later this month. Should we have two flags then, one for each, or what do you think?

trim21:
> I must be wrong then, well, in that case it should be possible given that we import a yaml parser. If you think yaml support is important I can try adding it later this month. Should we have two flags then, one for each, or what do you think?

can't parseFn parse file content by file name ext？

rasmus-kirk:
If we add a wrapper function that reads the file extension and applies the correct parsing function accordingly, then yeah I think so. Should be cleaner.

rasmus-kirk:
Because I do not have interest in working on it anymore, since it's not an issue for me personally anymore. If anyone else would pick it up it could be merged I guess.

PR #768, opened 2024-07-22
fix(qBittorrent): Add torrent files without authentication

almarzn:
Fixes add torrent files with no authentication using qbittorrent.

## Description

authCookie resolves to undefined when used without authentication. This lead to axios throwing ERR_HTTP_INVALID_HEADER_VALUE. Fixed to use the same method getRequestHeaders instead of the direct this.authCookie call, so it does not try to send the request with Cookie: undefined header.

## Types of changes


trim21:
thanks

Issue #605, opened 2022-10-05
Can only add tags in flood (torrent client is qBittorrent)


runaroundhome:
Type: Bug Report

## Your Environment
Running your docker container `jesec/flood:master`

- Version used:
  - Commit ID (development release): 6ff622c6b98fc896fcbd0657691e2a4a3b746d3b
- Environment name and version:
  - Node.js version: v18.9.0
  - npm version: 8.19.1
  - Web browser: Brave v1.44.105
- Operating system and version: Alpine 3.16.2 (in docker container)
- Torrent client and version: qBittorrent 4.4.5

## Summary
Removing tags doesn't work. (can only add tags)

Have also tested removing tags on qbt's native web ui, which works (and then does update in flood). But I can't remove tags in the flood UI.

Looking in the browser console it reports status 200 for the `api/torrents/tags` endpoint when removing a tag, but nothing is removed. The payload also looks correct, with the torrent hash and `tags: []`

## Expected Behavior
The tag should be removed.

## Current Behavior
Tags never get removed, can only add.

## Possible Solution
-

## Steps to Reproduce
1. Select "Set Tags" on a torrent
2. Delete all the tags and replace it with something new
3. Click "Set Tags" button (tag gets added but nothing removed)

Same goes if I "Set Tags" and remove a single or all tags, and click "Set Tags", nothing happens. (I mean the console says ok, 200, but the tags are still there)

trim21:
please try master branch when ci is completed

SDS1337:
Works!

PR #770, opened 2024-08-09
Fix qBittorrent tag removal

LXGaming:
## Description
Flood is attempting to remove the tag `undefined` when it's suppose to be removing all tags.

## Related Issue
#605 

## Types of changes
Coalesce `undefined` to a blank string.


trim21:
so `{tags: ''}` will remove all tags?

LXGaming:
Sorry I should have mentioned that in the PR, Yes sending `{tags: ''}` will remove all tags, this behavior is documented
<https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-4.1)#remove-torrent-tags>
> `tags` is the list of tags you want to remove from passed torrents. Empty list removes all tags from relevant torrents.

PR #776, opened 2024-08-09
build!: remove deprecated pkg and use nodejs SEA

trim21:
## Description

remove `vercel/pkg` and use nodejs [sea](https://nodejs.org/api/single-executable-applications.html)

This is a breaking change because it now require nodejs>=20

## Types of changes

- [x] Breaking change (changes that break backward compatibility of public API or CLI - semver MAJOR)


trim21:
closed as https://github.com/jesec/flood/pull/868

PR #779, opened 2024-08-10
Add a user agent for rss

zoriya:
## Description

Adds a user agent to the http call used to retrive rss feeds. This is needed because some rss servers refuses to answer if no user agent is set.

## Types of changes

### DIFF
--- a/server/models/FeedReader.ts
+++ b/server/models/FeedReader.ts
@@ -54,6 +54,11 @@ class FeedReader {
       maxHistory: this.options.maxHistory,
       interval: this.options.interval,
       forceInterval: true,
+      requestOpts: {
+        headers: {
+          'User-Agent': 'flood',
+        },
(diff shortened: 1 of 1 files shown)


trim21:
use `requestOpts` of feedsub please.

zoriya:
If you wonder why the header is assigned via Miniget's default option instead of feedsub it's because this option is not exposed via feedsub.

I didn't want to PR feedsub to expose this option so I used the global miniget option instead (also feedsub seems abandonned so i'm not sure if it would have been merged)

trim21:
> this option is not exposed via feedsub.

It's exposed

image
image

zoriya:
I remember looking at the source code because the user agent was overridden, but I don't remember that well & it doesn't seem to be the case looking at the code. I updated this, mb

Issue #720, opened 2024-02-20
transmission: -1 ratio on seeding torrents with 0B downloaded


antifuchs:
Type: Bug Report

- [x] Try to follow the update procedure described in the README and try again before opening this issue.

## Your Environment

- Version used: (via [nixos](https://github.com/NixOS/nixpkgs/blob/nixos-23.11/pkgs/applications/networking/p2p/flood/default.nix))
  - Version (stable release) 4.7.0
  - Commit ID (development release) `git --no-pager log -1`: 2b652f8148dab7134eeeb201b9d81dd6b8bda074
- Environment name and version:
  - Node.js version `node --version`
  - npm version `npm --version`
  - Web browser `name and version`
- Operating system and version: NixOS 23.11
- Torrent client and version: Transmission 3.00

## Summary

I'm not sure why the transmission API does this, but on occasion some seeding torrents will show 0 bytes downloaded; that causes flood to display the up/download ratio as `-1`.

## Expected Behavior

On torrents that were downloaded from elsewhere, it would be great if the ratio could be correctly computed, instead of punting on the computation. The transmission API reports an `uploadRatio` field (in the [RPC spec for `torrent-get`](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#33-torrent-accessor-torrent-get)), and it tends to be correct. Maybe flood can use that?

If that's not usable, I would suggest using the `totalSize` field as the divisor instead of `downloadedEver` (as total size tends to not be zero).

## Current Behavior

Currently, it looks like flood is dividing the amount of data uploaded (`uploadedEver`) by the amount of data downloaded (`downloadedEver`). If the divisor should be zero, that results in a not-a-number result, which I assume flood translates into `-1`.

I imagine it is reasonable for torrents that this instance started seeding, but transmission also reports 0B total downloaded on torrents that it downloaded from somewhere else (again, not sure why).

## Possible Solution

Several options:
* use `uploadRatio` from the transmission RPC API
* use `totalSize` (which is accurately reported by the API) instead of `downloadedEver` (which isn't) - but only if the torrent didn't start seeded... I imagine there would be a difference in the "date added" and "date finished" fields that could be used.

## Steps to Reproduce

I could not find a common pattern for when the transmission API reports useless field values (it has similar issues with the "date finished" field on occasion), sorry /:

trim21:
how you define "correct" share ratio if you didn't download anything...

antifuchs:
[...]

PR #791, opened 2024-08-20
changed transmission ratio calculation to total size instead of downloaded size

RA341:
Transmission ratio calculation was based on ```torrent.uploadedEver/torrent.downloadedEver```, which caused some torrents not initiated in flood to have a ratio of -1.

## Description

The calculation is now based on ```torrent.uploadedEver / torrent.totalSize```

## Related Issue
#720

## Screenshots

Before change:

Note the ratio is -1

image

After change

Note the ratio is now correctly displayed: 0.01

image

### DIFF
--- a/server/services/Transmission/clientGatewayService.ts
+++ b/server/services/Transmission/clientGatewayService.ts
@@ -383,7 +383,10 @@ class TransmissionClientGatewayService extends ClientGatewayService {
           ...(await Promise.all(
             torrents.map(async (torrent) => {
               const percentComplete = (torrent.haveValid / torrent.totalSize) * 100;
-              const ratio = torrent.totalSize === 0 ? -1 : torrent.uploadedEver / torrent.totalSize;
+              const ratio =
+                torrent.downloadedEver === 0
+                  ? torrent.uploadedEver / torrent.totalSize
+                  : torrent.uploadedEver / torrent.downloadedEver;
               const trackerURIs = getDomainsFromURLs(torrent.trackers.map((tracker) => tracker.announce));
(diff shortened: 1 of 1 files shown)


server/services/Transmission/clientGatewayService.ts
```diff
@@ -383,7 +383,7 @@ class TransmissionClientGatewayService extends ClientGatewayService {
           ...(await Promise.all(
             torrents.map(async (torrent) => {
               const percentComplete = (torrent.haveValid / torrent.totalSize) * 100;
-              const ratio = torrent.downloadedEver === 0 ? -1 : torrent.uploadedEver / torrent.downloadedEver;
+              const ratio = torrent.totalSize === 0 ? -1 : torrent.uploadedEver / torrent.totalSize;
```
trim21, line 386: how could `torrent.totalSize === 0` ?

RA341: It's should not happen, but if it does might as well handle it.

trim21: Don't think this will happen. also it's wrong. we should use totalSize based ratio only when torrent is not downloaded at all, which means `torrent.downloadedEver === 0`, not `torrent.totalSize === 0`

RA341: I see so something like this  ?

```ts
const ratio = torrent.downloadedEver === 0 ? -1 torrent.uploadedEver / torrent.totalSize : torrent.uploadedEver / torrent.downloadedEver;
```

trim21: yes

PR #792, opened 2024-08-21
Exatorrent support

Nathan-Kr:
Add support for [exatorrent](https://github.com/varbhat/exatorrent)

Works from version [sha-d2b6fdf](https://github.com/varbhat/exatorrent/pkgs/container/exatorrent/274474518?tag=sha-d2b6fdf)

## Description
supported:
- torrent list
- torrent management (add, start, stop, delete)

not supported:
- settings (no api to tweak settings)
- tags (not supported by exatorrent)
- path change (exatorrent use fixed path)
- priority (not supported by exatorrent)


trim21:
will you maintain it in the future? I don't use it and I dont' want to add a new client that no one maintaining

Nathan-Kr:
I'll keep it in a fork as I would anyway be the only one using it

trim21:
make sence, also save both time for and code review process.

trim21:
I think you can add your fork to readme about Exatorrent.

https://github.com/jesec/flood?tab=readme-ov-file#supported-clients

PR #810, opened 2024-10-09
fix: add_paused changed to add_stopped in qbit 5

whiteout12:
## Description
In Qbittorrent 5 and later the "add_paused" option is changed to "add_stopped". This fix checks the apiversion and uses the correct key.

## Types of changes
Dynamically set add_paused or add_stopped depending on apiversion

### DIFF
--- a/server/services/qBittorrent/types/QBittorrentTorrentsMethods.ts
+++ b/server/services/qBittorrent/types/QBittorrentTorrentsMethods.ts
@@ -123,6 +123,8 @@ export interface QBittorrentTorrentsAddOptions {
   skip_checking?: boolean;
   // Add torrents in the paused state. Possible values are true, false (default)
   paused?: boolean;
+  // Add torrents in the stopped state (using webapiVersion v2.11.0 or later). Possible values are true, false (default)
+  stopped?: boolean;
   // Create the root folder. Possible values are true, false, unset (default)
   root_folder?: boolean;
   // Content layout mode, replaces root_folder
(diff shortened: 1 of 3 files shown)


server/services/qBittorrent/types/QBittorrentTorrentsMethods.ts
```diff
@@ -123,6 +123,8 @@ export interface QBittorrentTorrentsAddOptions {
   skip_checking?: boolean;
   // Add torrents in the paused state. Possible values are true, false (default)
   paused?: boolean;
+  // Add torrents in the stopped state (using webapiVersion v2.11.0 or later). Possible values are true, false (default)
+  stopped?: boolean;
```
trim21, line 127: isn't “add stop” has same meaning as ”add paused“？

whiteout12: It's to take care of the api changes with qbit 5 where it's now using "stopped" instead of "paused". So in order to fit both earlier and later versions of the api both should be there. Makes sense?

trim21: I understand there is a api change in qb v5, but in my in my understanding it rename from "paused" to "stopped", and they have actually same meaning, correct?

So we I think we do not need to handle it outside clientGatewayService.ts and clientRequestManager.ts

whiteout12: Forgive me if im wrong but doesn't "stopped" need to be a valid key when creating the options (based on QBittorrentTorrentsAddOptions interface) in ` await this.clientRequestManager.torrentsAddFiles(fileBuffers, {.....` line 88 in clientGateWayService ?

trim21: OK, I understand what you are doing here.

Issue #820, opened 2024-11-14
flood with qbittorrent : No such method TorrentsController


Kosteron:
Type: Bug Report

- [ ] Try to follow the update procedure described in the README and try again before opening this issue.

## Your Environment

- Version used:
  - Version : Flood server 4.8.4
  - Commit ID : none
- Environment name and version:
  - Node.js version `the one from the flood docker`
  - npm version `the one from the flood docker`
  - Web browser : Firefox 128.3.1esr (64 bits)
- Operating system and version: Linux
- Torrent client and version: QBITTORRENT_VERSION=5.0.1-r0 as mentioned [here](https://hub.docker.com/layers/linuxserver/qbittorrent/latest/images/sha256-25127db73150b3de5134e35849ac436136cdf711b6b1750e35cf1060d2d27168?context=explore)

## Summary

Hi, I'm trying to create a docker-compose stack with flood as front-end for qbittorrent, but I have some trouble with it.
Once everything is started, I can add a torrent, but when I want to stop or resume it, I got this error

``` bash
...
qbittorrent          | QMetaObject::invokeMethod: No such method TorrentsController::pauseAction()
flood                | POST /api/torrents/stop 500 17.917 ms - 74
qbittorrent          | QMetaObject::invokeMethod: No such method TorrentsController::pauseAction()
flood                | POST /api/torrents/stop 500 11.144 ms - 74
qbittorrent          | QMetaObject::invokeMethod: No such method TorrentsController::resumeAction()
flood                | POST /api/torrents/start 500 15.480 ms - 74
...
```

Also the flood UI blink and show shortly the "unable to connect to qbittorrent" UI and go back to the main page

## Expected Behavior

The torrent should be stopped or started accordingly

## Current Behavior

Unable to start or stop torrent

## Possible Solution

Maybe update the method call by flood for latest qbittorrent version ?

## Steps to Reproduce

1. Start a docker-compose stack with jesec/flood latest version and lscr.io/linuxserver/qbittorrent latest version
2. Add a torrent on flood and stop it
3. See the error 500 message on the output of docker compose log

Here is a sample docker-compose file to reproduce the problem :

``` yaml
services:
  flood:
    hostname: flood
    container_name: flood
    image: jesec/flood
    user: 1001:1001
    restart: unless-stopped
    command: --port 3001
    ports:
      - 3001:3001
    environment:
      - PUID=1001
      - PGID=1001
      - HOME=/config
    volumes:
      - qbittorrent-config:/config
[...]

PR #826, opened 2024-11-25
fix(qBittorrent): API Version state handling

LXGaming:
## Description
When Flood is started **before** a qBittorrent 5 client, starting and stopping torrents will fail due to the wrong method being called.

This is happening due to Flood getting the API Version in the clientRequestManager constructor ([clientRequestManager.ts#L626](https://github.com/jesec/flood/blob/843b9c966cabcbcc542210ee0d08c4262255278f/server/services/qBittorrent/clientRequestManager.ts#L626)) and nowhere else. If this request fails then the `apiVersion` will remain null until Flood is restarted.

I made the `apiVersion` consistent with the `authCookie` so that it's kept in sync with the qBittorrent client and can be awaited to ensure the value is resolved. I also renamed `updateAuthCookie` to `updateConnection` so that it better represents what the method is now doing.

As `getApiVersion` requires authentication I've chained the promises so that it runs only when `authenticate` is successful.

The changes to `clientGatewayService` are to make the calls to `isApiVersionAtLeast` consistent with `clientRequestManager`.

## Related Issue
#820 

## Types of changes

### DIFF
--- a/server/services/qBittorrent/clientRequestManager.ts
+++ b/server/services/qBittorrent/clientRequestManager.ts
@@ -36,7 +36,7 @@ const EMPTY_SERVER_STATE = {
 class ClientRequestManager {
   private connectionSettings: QBittorrentConnectionSettings;
   private apiBase: string;
-  apiVersion: string | null = null;
+  apiVersion: Promise<string | undefined> = Promise.resolve(undefined);
   private authCookie: Promise<string | undefined> = Promise.resolve(undefined);
   private isMainDataPending = false;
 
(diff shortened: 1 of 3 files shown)


server/services/qBittorrent/clientRequestManager.ts
```diff
@@ -36,7 +36,7 @@ const EMPTY_SERVER_STATE = {
 class ClientRequestManager {
   private connectionSettings: QBittorrentConnectionSettings;
   private apiBase: string;
-  apiVersion: string | null = null;
+  apiVersion: Promise<string | undefined> = Promise.resolve(undefined);
```
trim21, line 39: why Promise here?

trim21: Ok I got it

LXGaming:
Force push to correct the commit summary and to resolve the linting fail.

server/services/qBittorrent/clientRequestManager.ts
LXGaming: Yeah I'm not sure if there is a better way to handle it, Having it update along side the `authCookie` seems like the easiest way to do it, and it needs to be await-able so that starting / stopping doesn't resolve to undefined before the request has completed.

PR #906, opened 2025-08-17
feat: Add optional Last Activity column to torrent list

raitono:
# Optional Last Activity column

## Description

This PR adds an optional Last Activity column to the torrent list. The Last Activity column displays when each torrent was last active, providing users with better visibility into their torrent activity over time.

The implementation includes:

- Adding the Last Activity column to the torrent list columns configuration
- Implementing the column display logic in TorrentListCell
- Adding sorting support for the new column
- Including internationalization strings for the column header
- Setting default preferences for the new column

## Related Issue

N/A

## Screenshots

image

image

## Types of changes


jesec:
Thanks for the contribution.

Keep in mind that raw data for those columns has to come from somewhere (either fetching from torrent clients i.e rTorrent or from Flood-internal bookkeeping).

I don’t see any source for this added information yet.

raitono:
This is an existing field that can already be found in the torrent details. It just didn't have a column. I'll try to check all the clients to make sure they have a data source.

PR #940, opened 2025-11-13
chore: upgrade create-torrent

euri10:
## Description

current version of create-torrent doesnt seem to limit the torrent piece size which is problematic as certain trackers will refuse the created .torrent

latest version of create-torrent defaults the max piece size of a torrent to 4MB

a lot of sites prevent having a piece size above 8MB so its good to have that

## Types of changes


trim21:
use pnpm please

jesec:
@claude review, fix and validate the PR. and maybe we should have more options in “Create Torrent” panel.

euri10:
1st time i hear about pnpm, hopefully i get this right, if not claude will ;)

trim21:
looks like there are some breaking changes

trim21:
looks like we will need to migrate from jest to vitest so we can use this esm only package in testing.

trim21:
I'll fix it later, close it for now

PR #967, opened 2026-01-02
[draft] Add the user name as tag when adding torrents (also when using "basic auth")

arximboldi:
## Description

This is probably not ready for everyone, but something I've done for myself.

I manage a seedbox that I share with a group of friends. It is hidden using "HTTP basic auth". It would be nice to be able to see who is downloading what ("is this stuff good?", "can we delete this to make space?", etc.). My solution here is to automatically add the current user to the list of tags.

There are two parts here. 

1) Because I am using Basic Auth (managed by an Nginx in front) I am reusing the `/verify` command to request the name of the of the user (one can't otherwise know directly from JS).

2) The actual auto-tag the user feature is added.

Do you think this feature could be useful? Should this behavior be hidden behind a command line argument, or maybe something in the settings dialog? Are there issues with the implementation, perhaps there is a cleaner way to deal with the basic-auth-via-nginx use-case?

## Types of changes

### DIFF
--- a/client/src/javascript/components/modals/add-torrents-modal/AddTorrentsByCreation.tsx
+++ b/client/src/javascript/components/modals/add-torrents-modal/AddTorrentsByCreation.tsx
@@ -1,4 +1,4 @@
-import {FC, use, useRef, useState} from 'react';
+import {FC, useRef, useState} from 'react';
 import {useLingui} from '@lingui/react';
 
 import {Checkbox, Form, FormHandle, FormRow, Textbox} from '@client/ui';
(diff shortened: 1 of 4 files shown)


trim21:
sorry I'm going to close this.

It would be a good idea that we have a plugin system that users could extends the flood as they need, for example, when adding torrents, a plugin know current users and add tag to new downlaods

But it's a bad idea users adding very narrowed feature to the codespace.

trim21:
also please don't commit nix files into repo

arximboldi:
Sure, no problem. Thanks for looking into it!

PR #1038, opened 2026-02-15
feat: support searching torrents by info hash

mmgoodnow:
this PR adds info hashes to the search term for each torrent that the search query matches against. The matching is already case-insensitive so it supports both uppercase info hashes and lowercase info hashes.


mmgoodnow:
@trim21 since it seems you are active, wanted to give you a gentle heads up to this one liner QOL improvement!

trim21:
I'm ok with the idea but I don't like how you implement it, we should do a fuzzy matching on name and no fuzzy matching on infohash

mmgoodnow:
I changed it to filter based on `search query fuzzy matches name OR search query "exactly" matches info hash (trimmed and case insensitive)`

trim21:
can you add some comment about the expected behavior?

PR #1048, opened 2026-02-22
fix(auth): enforce case-insensitive username checks across auth and self-deletion flows

marcinmajsc:
## Description
I’m not sure whether this is a bug or intended behavior, but the login form is case-sensitive with regard to the username.
For me (and most likely for others as well), this is very annoying, as I’ve often instinctively used a different letter case for the username than the one used during registration, which resulted in the authentication being rejected.

This PR improves username handling in authentication and authorization flows by enforcing case-insensitive behavior end to end.

### What changed
- Added case-insensitive username uniqueness checks when creating users, so usernames differing only by letter case cannot coexist.
- Updated login-related user lookups to treat usernames case-insensitively.
- Updated delete-user safeguards to block self-deletion even when username casing differs (for example, `Admin` vs `admin`).
- Updated the Settings/Auth tab client behavior to align self-delete checks with the same case-insensitive logic.
- Added/extended API tests to cover case-insensitive create, login, and self-delete scenarios.

### Why
These changes close a class of edge cases where inconsistent casing could bypass duplicate-user protections or self-deletion safeguards, improving account integrity and preventing accidental or unsafe account operations.

## Related Issue
N/A

## Types of changes

### DIFF
--- a/client/src/javascript/components/modals/settings-modal/AuthTab.tsx
+++ b/client/src/javascript/components/modals/settings-modal/AuthTab.tsx
@@ -125,7 +125,7 @@ const AuthTab: FC = observer(() => {
               .slice()
               .sort((a: Credentials, b: Credentials) => a.username.localeCompare(b.username))
               .map((user: Credentials) => {
-                const isCurrentUser = user.username === AuthStore.currentUser.username;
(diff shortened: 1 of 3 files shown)


trim21:
This looks like s bad idea for me

marcinmajsc:
> This looks like s bad idea for me

Sure, I understand :) In that case, the PR can be closed.

Issue #631, opened 2023-01-29
Empty torrent list when connected to Deluge


josh-gaby:
Type: Bug Report

## Your Environment

- Version used: 
  - Version (stable release)  `4.7.0` via `Docker figro/flood-ui`
- Environment name and version:
  - Node.js version `19.4.0`
  - npm version `9.2.0`
  - Web browser `Chrome 109.0.5414.87`
- Operating system and version: Accessing on `MacOS 13.1`, Dockers running on `Unraid 6.11.5` 
- Torrent client and version: `Deluge 2.1.1` via `binhex/arch-delugevpn`

## Summary

Deluge 2.1.1 running in Docker on Unraid, Flood running in Docker on Unraid No torrents listed in UI but transfer rates updating.

![image](https://user-images.githubusercontent.com/28797673/215324201-740232a1-7832-4bec-9b38-957cf4d11cbe.png)

## Expected Behavior

List of torrents should be populated.

## Current Behavior

No torrents are being displayed, the EventStream is returning an empty object for TORRENT_LIST_FULL_UPDATE

![image](https://user-images.githubusercontent.com/28797673/215324384-2d2297e9-14b8-4bed-8b64-5242992ab85c.png)

## Steps to Reproduce

1. Run Deluge in Docker
2. Run Flood in Docker
3. Connect Flood to Deluge

mkataja:
I've got a similar issue with 2 of my 4 rtorrent instances. I can't figure out any obvious difference between the ones that work and the ones that don't.

awsms:
> I've got a similar issue with 2 of my 4 rtorrent instances. I can't figure out any obvious difference between the ones that work and the ones that don't.

Can you try out https://github.com/jesec/flood/pull/1098 ? I was having the same issue for a while and it fixed it.

mkataja:
> Can you try out [#1098](https://github.com/jesec/flood/pull/1098) ? I was having the same issue for a while and it fixed it.

I didn't try your PR, but yes, my issue was about JSON RPC too, and falling back to XML RPC would avoid the error. So your PR would most likely solve this for me. The underlying rtorrent issue is tracked here: https://github.com/rakshasa/rtorrent/issues/1596

PR #1098, opened 2026-04-17
fix(rtorrent): fall back to XML-RPC for torrent list fetch

awsms:
## Description

Some rTorrent instances return JSON-RPC parse errors (-32700) for the d.multicall2 torrent list request, even though other RPC calls still succeed. This caused Flood to silently fall back to an empty torrent list while transfer stats and add-torrent operations kept working.

Retry the torrent list fetch over XML-RPC when that specific JSON-RPC failure occurs, keeping the existing behavior for all other calls.

## Types of changes


trim21:
I don't quiet understand, you are linking this PR in https://github.com/jesec/flood/issues/631,  are we talking about Deluge or rtorrent?

trim21:
OK, I know.

I need you to add this pr description as comment to the code so we don't need to use `git blame` to understand why we are doing this.

awsms:
> I need you to add this pr description as comment to the code so we don't need to use `git blame` to understand why we are doing this.

Done, I've also referenced the rTorrent issue #.

PR #1121, opened 2026-05-11
fix: filter qBittorrent trackers by tier>=0 for message extraction

### LINKED ISSUE #1118: Incorrect logic for torrent tracker error mesasges
Type: Bug Report

The most recent update (4.14.0) treats any tracker message as a warning.

I only use private trackers and all of them have messages that state "This torrent is private".

It is not a warning level message so I think the logic might need refining.

Perhaps there should be a toggle to enable or disable error state for tracker messages.

How UI looks in Flood:
Image

What the tracker messages are in qBit:
Image

trim21:
## Summary

- Filter qBittorrent trackers by `tier >= 0` when extracting tracker messages and building tracker URIs
- DHT/PeX/LSD trackers (tier < 0) send informational messages like "This torrent is private" that were incorrectly triggering the warning status introduced in #1108
- `isPrivate` detection still uses the raw tracker list (index 0) since DHT trackers may carry the "is private" message

Fixes #1118

### DIFF
--- a/server/services/qBittorrent/clientGatewayService.ts
+++ b/server/services/qBittorrent/clientGatewayService.ts
@@ -368,17 +368,14 @@ class QBittorrentClientGatewayService extends ClientGatewayService {
             const trackers = await this.clientRequestManager.getTorrentTrackers(hash).catch(() => undefined);
 
             if (properties != null && trackers != null && Array.isArray(trackers)) {
-              const firstTrackerMsg = trackers.find((t) => t.msg.length > 0)?.msg ?? '';
+              const realTrackers = trackers.filter((t) => t.tier >= 0);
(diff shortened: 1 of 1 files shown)


zotabee:
Does this fix the following problem?:

I updated too soon to a new version of qBittorrent and some of my private trackers returned a "banned client" message (Tracker Message : Banned client, please see the rules for a list of allowed clients).

Few days later, client is whitelisted, qBittorrent WebUI shows the current torrent status, "Working" in green, no more tracker errors.

But in flood, these torrents are still in orange with the original "Tracker Message : Banned client" message.

trim21:
looks like bug because we cache torrent properties forever, I'll fix this later

Issue #1117, opened 2026-05-10
qBittorrent 5.2 support


thefat32:
Type: Bug Report

- [x] Try to follow the update procedure described in the README and try again before opening this issue.

## Summary

It seems like updating to qBittorrent 5.2 breaks flood connection with Web API

There has been some changes in the release

<img width="589" height="212" alt="Image" src="https://github.com/user-attachments/assets/b17165e1-5bdd-48c5-8004-4b5743288fe7" />

https://www.qbittorrent.org/news#sun-may-03rd-2026---qbittorrent-v5.2.0-release

It would be great to support it or at least mark as not supported in the README

trim21:
Flood has added support for qB 5.2 in the latest update. If you run into any issues, please provide a description of the bug.

thefat32:
Thanks @LXGaming I did not have the time to track the issue down to the cause. Thanks also @trim21 for your answer and support.

PR #1124, opened 2026-05-15
fix(qBittorrent): session cookie name

LXGaming:
## Description
After updating to qBittorrent 5.2, Flood is failing to connect with the error below, I found that the session cookie name had been changed from `SID` to `QBT_SID_<WEB_UI_PORT>` (For example `QBT_SID_8080`) https://github.com/qbittorrent/qBittorrent/commit/753fb80e9bf4aa96dbb929f16d29c0156069a2c9
```
{
    "code": "INTERNAL_SERVER_ERROR",
    "message": "AxiosError: Request failed with status code 403"
}
```

## Related Issue
- https://github.com/jesec/flood/issues/1117

## Types of changes
I added a check for `QBT_SID_`  along side the check for `SID=` to support qBittorrent 5.2 and maintain backwards compatibility.

### DIFF
--- a/server/services/qBittorrent/clientRequestManager.ts
+++ b/server/services/qBittorrent/clientRequestManager.ts
@@ -73,7 +73,7 @@ class ClientRequestManager {
         const cookies = res.headers['set-cookie'];
 
         if (Array.isArray(cookies)) {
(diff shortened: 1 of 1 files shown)


trim21:
I'm thinking maybe we should simply preserve all cookies?

PR #1132, opened 2026-05-16
fix(qBittorrent): add 1h TTL for cached tracker messages and URIs

trim21:
The qBittorrent adapter caches tracker messages and URIs indefinitely after the first fetch. This means tracker messages that change over time (e.g. Banned client to Working after client whitelisting) are never refreshed in Flood, even though qBittorrent WebUI shows the updated status.

Changes:
- Extracted fetchProperties() method for reusing the cache population logic
- Added trackerMessageFetchedAt timestamp to cached properties
- Added 1-hour TTL: when the cached tracker data exceeds 1 hour, the tracker endpoints are re-queried and both trackerMessage and trackerURIs are refreshed

Fixes https://github.com/jesec/flood/pull/1121#issuecomment-4466630056

### DIFF
--- a/server/services/qBittorrent/clientGatewayService.ts
+++ b/server/services/qBittorrent/clientGatewayService.ts
@@ -51,8 +51,41 @@ class QBittorrentClientGatewayService extends ClientGatewayService {
     string,
     Pick<TorrentProperties, 'comment' | 'dateCreated' | 'isPrivate' | 'trackerURIs'> & {
       trackerMessage: string;
+      trackerMessageFetchedAt: number;
     }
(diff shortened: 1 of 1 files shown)


zotabee:
Thank you! So fast!

runaroundhome:
Thank you soooo much for fixing this! It's been a bug in the OG flood back in the day. I think I reported it but was never fixed, that is, till today. So again, huge thank you.

I can finally put to bed my workaround I had added. A script that manually checked everything day, in qbt and then adding a tag to flood. It wasn't the best, but it worked.

PR #1135, opened 2026-05-17
chore(Dockerfile): set source to the GH repository

slagiewka:
## Description
This should allow tools (e.g. Renovate in my case) to find their way back to where the image originates from.

I'm expecting Renovate to crawl back to the repository and fetch the changelog for me this way.

## Types of changes

### DIFF
--- a/distribution/containers/Dockerfile.distroless
+++ b/distribution/containers/Dockerfile.distroless
@@ -28,6 +28,7 @@ RUN chown 1001:1001 /root/sysroot/home/download
 
 # flood image
 FROM scratch AS flood
+LABEL org.opencontainers.image.source="https://github.com/jesec/flood"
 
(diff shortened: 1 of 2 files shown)


trim21:
actually https://github.com/jesec/flood/tree/master/distribution/containers

slagiewka:
> actually https://github.com/jesec/flood/tree/master/distribution/containers

That' so silly of me. Fixed.

I didn't touch `Dockerfile.rtorrent` as I'm not sure where that should point in the end.

trim21:
> > actually https://github.com/jesec/flood/tree/master/distribution/containers
> 
> That' so silly of me. Fixed.
> 
> I didn't touch `Dockerfile.rtorrent` as I'm not sure where that should point in the end.

yeah maybe we should drop it from this repo

PR #1004, opened 2026-01-27
feat: add configurable bottom panel for torrent details

mReXiTuS:
Add a persistent, resizable bottom panel to display torrent details as an alternative to the modal popup. The panel can be enabled/disabled via user settings and includes the following features:

Features:
- Persistent bottom panel showing torrent details (Details, Files, Peers, Trackers, MediaInfo tabs)
- Vertically resizable with height persistence (200px min, viewport-200px max)
- Configurable via Settings > User Interface > "Show details as bottom panel"
- Single-click shows panel on desktop (when enabled)
- ESC key closes panel
- Mobile devices continue using modal (responsive design)
- Panel background matches torrent list for visual consistency

Implementation:
- Add UIStore.detailsPanelHash and detailsPanelVisible state management
- Add FloodSettings.UITorrentDetailsPanel and detailsPanelHeight settings
- Create TorrentDetailsPanel and TorrentDetailsPanelResizeHandle components
- Update modal tab components to support dual-source (panel or modal)
- Add settings UI in UITab for panel configuration
- Update click handlers to respect panel setting
- Add CSS Grid layout for panel integration

Breaking Changes: None
- Default behavior shows panel (enabled by default)
- Users can disable to restore original modal-only behavior
- Mobile behavior unchanged (always uses modal)

## Screenshots

Screenshot 2026-01-27 at 18 23 35

PR #1150, opened 2026-06-11
feat: add configurable bottom panel for torrent details v2

marcinmajsc:
## Description
This pull request is based on #1004 with changes:
- Merge commits from master
- Fixed dependency issues
- Improved window display
- Improved file list display
- Deleted newly created settings group
- Moved options to the current group

## Screenshots
image image 

## Types of changes

- [X] Breaking change (changes that break backward compatibility of public API or CLI - semver MAJOR)

Related to #1004


trim21:
OK I'm not sure how should I handle this PR. the owner of https://github.com/jesec/flood/pull/1004 didn't response for a long time.

@mReXiTuS any thought? are you going to complet your PR?

marcinmajsc:
@trim21 his work has been included. This PR contains the original commits with its changes.

trim21:
> @trim21 his work has been included. This PR contains the original commits with its changes.

Do you mind I wait for his reponse for a week? then I'll review your PR

marcinmajsc:
> > @trim21 his work has been included. This PR contains the original commits with its changes.
> 
> Do you mind I wait for his reponse for a week? then I'll review your PR

Sure 😊

client/src/javascript/components/torrent-details-panel/TorrentDetailsPanelResizeHandle.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    setIsDragging(false);
+
+    if (resizeLine.current != null) {
+      resizeLine.current.style.opacity = '0';
+      resizeLine.current.style.transform = '';
+    }
+
+    if (startY.current != null && initialHeight.current != null) {
+      const deltaY = startY.current - (window.event as PointerEvent).clientY;
```
trim21: `window.event` is already deprecated

trim21:
I feel like this doesn't look much like React code — isn't there too much direct DOM manipulation?

marcinmajsc:
> I feel like this doesn't look much like React code — isn't there too much direct DOM manipulation?

@trim21 I corrected the code. Is it ok now?

trim21 (changes requested):
## Review Summary

Thanks for picking up #1004 and improving it! The refactored resize handle is much cleaner. Overall direction is good, but there are a few issues to address before merging.

### Critical

1. **Height source conflict in `TorrentDetailsPanel`** — reads stale height from SettingStore during drag (see inline comment)

### Should Fix

2. **Unnecessary `observer` HOC** in `TorrentDetailsPanelResizeHandle` — component reads no observables
3. **Panel fallback to modal** when `UITorrentDetailsPanel` is disabled — double-click currently does nothing
4. **Window resize handling** — panel height not re-constrained when browser window shrinks
5. **CSS transition during drag** — `transition: grid-template-rows 0.2s ease` fights with real-time drag updates

### Nit

6. **Missing tests** — no Storybook stories for the new panel components
7. **ESC key cleanup** — TorrentDetailsPanel always adds a keydown listener even when hidden

client/src/javascript/components/torrent-details-panel/TorrentDetailsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+      content: TorrentTrackers,
+    },
+    {
+      id: 'torrent-mediainfo',
+      label: i18n._('torrents.details.mediainfo'),
+      content: TorrentMediainfo,
+    },
+  ];
+
+  const height = SettingStore.floodSettings.detailsPanelHeight || 400;
```
trim21, line 61: **🔴 Issue: Height source conflict during drag**

During resize drag, `Overview.tsx` tracks height in React state (`panelHeight`) and applies it via CSS `--details-panel-height`, which updates the grid row size in real time. However, this line reads directly from `SettingStore.floodSettings.detailsPanelHeight`, which is only updated on `pointerup` (via `SettingActions.saveSetting`).

This means the `style={{height: ...}}` inline style on the panel div is stale during drag — it uses the pre-drag value while the grid row is already at the new size. The two fight each other.

**Fix**: Accept `height` as a prop from `Overview.tsx` and use it here instead of reading from SettingStore.

```tsx
// Add to props:
interface TorrentDetailsPanelProps { height?: number; }
// Use the prop:
const TorrentDetailsPanel: FC<TorrentDetailsPanelProps> = observer(({height: propHeight}) => {
  const height = propHeight ?? SettingStore.floodSettings.detailsPanelHeight ?? 400;
```
Then in `Overview.tsx`: `<TorrentDetailsPanel height={panelHeight} />`

client/src/sass/components/_torrent-details-panel-bottom.scss
```diff
@@ -0,0 +1,146 @@
+@use '../tools/themes';
+@use '../tools/variables';
+
+.view--torrent-list {
+  &.has-details-panel {
+    display: grid;
+    grid-template-rows: 1fr auto var(--details-panel-height, 400px);
+    transition: grid-template-rows 0.2s ease;
```
trim21, line 8: **🟡 Issue: `transition` conflicts with real-time drag**

`transition: grid-template-rows 0.2s ease` causes a 200ms animation delay when `--details-panel-height` changes. During drag, this CSS variable updates on every `pointermove` event (60+ times per second), but the transition makes the visual response lag behind the user's cursor.

**Fix**: Disable transition during drag. One approach — add a class like `.is-dragging` on the container and use:
```scss
&.has-details-panel {
  display: grid;
  grid-template-rows: 1fr auto var(--details-panel-height, 400px);
  &:not(.is-dragging) {
    transition: grid-template-rows 0.2s ease;
  }
}
```
Toggle the class from `TorrentDetailsPanelResizeHandle`'s drag state.

client/src/javascript/components/torrent-details-panel/TorrentDetailsPanelResizeHandle.tsx
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const getConstrainedHeight = (initialHeight: number, startY: number, currentY: number) => {
+  const deltaY = startY - currentY;
+  const maxHeight = window.innerHeight - WINDOW_PADDING;
+
+  return Math.max(MIN_PANEL_HEIGHT, Math.min(maxHeight, initialHeight + deltaY));
+};
+
+const TorrentDetailsPanelResizeHandle: FC<TorrentDetailsPanelResizeHandleProps> = observer(
```
trim21, line 27: **🟡 Issue: Unnecessary `observer` HOC**

This component does not read any MobX observables — it only uses React props (`height`, `onHeightChange`) and local state (`useState`, `useRef`). The `observer` wrapper adds unnecessary overhead (subscribing to observable changes that never happen).

**Fix**: Remove the `observer()` wrapper and the import:
```tsx
const TorrentDetailsPanelResizeHandle: FC<TorrentDetailsPanelResizeHandleProps> = ({height, onHeightChange}) => {
```

client/src/javascript/components/torrent-list/TorrentListRow.tsx
```diff
@@ -50,6 +51,17 @@ const displayContextMenu = (hash: string, event: KeyboardEvent | MouseEvent | To
...
 const displayTorrentDetails = (hash: string) => UIStore.setActiveModal({id: 'torrent-details', hash});
 
+const showTorrentInPanel = (hash: string) => {
+  const usePanelView = SettingStore.floodSettings.UITorrentDetailsPanel ?? true;
+  if (usePanelView) {
+    UIStore.setDetailsPanelHash(hash);
+    if (!UIStore.detailsPanelVisible) {
+      UIStore.setDetailsPanelVisible(true);
+    }
+  }
```
trim21, line 61: **🟡 Issue: No fallback to modal when panel is disabled**

When the user disables `UITorrentDetailsPanel` in settings, clicking/double-clicking a torrent row on desktop does nothing — the torrent gets selected but no detail view is shown at all. The original behavior (before this PR) always triggered `displayTorrentDetails(hash)` which opened the modal.

Users who prefer the modal over the panel will find double-click broken.

**Fix**: Add a fallback:
```tsx
const showTorrentInPanel = (hash: string) => {
  const usePanelView = SettingStore.floodSettings.UITorrentDetailsPanel ?? true;
  if (usePanelView) {
    UIStore.setDetailsPanelHash(hash);
    if (!UIStore.detailsPanelVisible) {
      UIStore.setDetailsPanelVisible(true);
    }
  } else {
    displayTorrentDetails(hash);
  }
};
```

client/src/javascript/components/torrent-details-panel/TorrentDetailsPanel.tsx
```diff
@@ -0,0 +1,117 @@
...
+import TorrentPeers from '../modals/torrent-details-modal/TorrentPeers';
+import TorrentTrackers from '../modals/torrent-details-modal/TorrentTrackers';
+
+const TorrentDetailsPanel: FC = observer(() => {
+  const {i18n} = useLingui();
+  const [activeTab, setActiveTab] = useState<string>('torrent-details');
+
+  const torrentHash = UIStore.detailsPanelHash;
+  const torrent = torrentHash ? TorrentStore.torrents[torrentHash] : undefined;
+
```
trim21, line 22: **🟡 Issue: ESC keydown listener always active even when panel is hidden**

This `useEffect` always registers a `keydown` listener on the window, even when the panel is not visible. Since `TorrentDetailsPanel` is always rendered (just conditionally displayed via `{showPanel && ...}` in `Overview.tsx`), the ESC handler runs even when the panel is closed.

In practice this is harmless (it just sets `detailsPanelVisible=false` which is already false), but it's extra overhead and a pattern concern. Consider either:
- Only rendering the component when `showPanel` is true, or
- Adding a condition inside the handler to check `UIStore.detailsPanelVisible` before firing.

trim21 (commented):
I'm not an expert on CSS or React, and this was reviewed with assistance from an LLM — feel free to tell me if I'm wrong about something.

trim21:
> > > @trim21 his work has been included. This PR contains the original commits with its changes.
> > 
> > 
> > Do you mind I wait for his reponse for a week? then I'll review your PR
> 
> Sure 😊

I hope we don't get another stale PR...

PR #1167, opened 2026-06-30
feat: register as OS handler for .torrent files and magnet links

shocklateboy92:
This makes Flood show up as an application to open torrent files with in OS native file type properties (when you install as PWA). With this, I can completely replace the Transmission desktop app with Flood :grin:.

## Description

Magnet links already worked via the in-app registerProtocolHandler button and the existing `?action=add-urls` query branch in AppWrapper; the manifest protocol_handlers entry just makes that registration happen automatically on install, reusing the same handler URL.

For .torrent files, I registered a `launchQueue` consumer (File Handling API) in TorrentListDropzone. The logic was the exact same as the drag-and-drop torrent-file intake, so I thought this was a good place - let me know if you want it somewhere else. `launchQueue` buffers the launch until the consumer mounts, so a cold launch still gets delivered after login.

While I was there, I noticed the base64 file processing was duplicated between `TorrentListDropzone` and `FileDropzone`. So I extracted it to `util/processFiles`.

I've tested and it works great on Chromium, but both handlers are feature-detected and should do nothing where the APIs are unavailable (Firefox, Safari/iOS).

## Screenshots

image

## Types of changes

### DIFF
--- a/client/src/javascript/components/general/form-elements/FileDropzone.tsx
+++ b/client/src/javascript/components/general/form-elements/FileDropzone.tsx
@@ -4,8 +4,9 @@ import {Trans} from '@lingui/react';
 
 import {Close, File, Files} from '@client/ui/icons';
 import {FormRowItem} from '@client/ui';
+import processFiles from '@client/util/processFiles';
 
-export type ProcessedFiles = Array<{name: string; data: string}>;
+import type {ProcessedFiles} from '@client/util/processFiles';
 
 interface FileDropzoneProps {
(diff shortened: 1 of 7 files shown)


trim21:
I'm no frontend expert, but I noticed you modified client/src/public/manifest.json. Does this mean this feature is only effective within the PWA?

shocklateboy92:
Yes, PWA only. Chromium only respects file handlers after you click "install as app". It would be chaos otherwise 😅

Issue #1275, opened 2026-07-15
TypeError [ERR_INVALID_ARG_VALUE] on master


zotabee:
Type: Bug Report

404 page not found

## Your Environment

Docker and master image

- Version used:

jesec/flood:master
Image : sha256:9e3c421f9b0db4a360d1760e67200cebdcec3157e54d868cb31f23b11a51a981

## Summary

```
Node.js v24.18.0

node:internal/modules/cjs/loader:2084

    throw new ERR_INVALID_ARG_VALUE('filename', filenameOrURL, createRequireError);

    ^

TypeError [ERR_INVALID_ARG_VALUE]: The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received undefined

    at createRequire (node:internal/modules/cjs/loader:2084:11)
    at Object.<anonymous> (/usr/local/lib/node_modules/@jesec/flood/dist/index.js:127644:88)
    at Module._compile (node:internal/modules/cjs/loader:1871:14)
    at Object..js (node:internal/modules/cjs/loader:2002:10)
    at Module.load (node:internal/modules/cjs/loader:1594:32)
    at Module._load (node:internal/modules/cjs/loader:1396:12)
    at wrapModuleLoad (node:internal/modules/cjs/loader:255:19)
    at Module.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:154:5)
    at node:internal/main/run_main_module:33:47 {

  code: 'ERR_INVALID_ARG_VALUE'

}
```

## Expected Behavior

Flood starts correctly

## Current Behavior

404 page not found

## Possible Solution

Switch to flood stable image

## Steps to Reproduce

1. try to open flood via the master image

PR #1325, opened 2026-07-20
build: replace esbuild with rolldown

slagiewka:
## Description

This moves the build step towards the broader Vite ecosystem. Since Vite already uses Rolldown under the hood, this might save some extra dependency in the futrue (esbuild is still pulled unfortunately).

Version 1.1.5 to match vite internal for the least pnpm-lock.yaml diff. I tested 1.2.0 and it has only minor differences in the output. However Vite pins to `~1.1.5`, so 1.2.0 is extra download.

The output index.js file goes down from 50M to 35M for ~30% reduction in size. Likely due to better tree-shaking(?). I did build the local Dockerfile and run it locally (w/o any torrent client connected) and it seemed to work fine. However, some more testing might be useful - not sure what else to check.

## Types of changes

I'm picking bug fix since it should theoretically be transparent to the end-user.


trim21:
I'm actually thinking about use vite directly since we already use vite-node to run development backend server

slagiewka:
> I'm actually thinking about use vite directly

I thought about it too. I'm no expert here, but I thought it might be a good first step before going into how can Vite be configured to achieve the same structure.

trim21:
the resize difference may be caused by different target node version, not tree-shake

slagiewka:
> the resize difference may be caused by different target node version, not tree-shake

Thanks for digging deeper. I initially couldn't find the target equivalent option. I didn't look thoroughly, though. 
Re-added and size is the same for me (still 35M).

trim21:
ok I had ai checked it again, the rolldown generate smaller source map

trim21:
I don't see any compelling reason to switch from esbuild to rolldown — the two bundlers seem comparable. That said, rolldown does produce smaller output, so it's not a downside either. However, there's a possible concern: esbuild currently struggles with yargs 18 and some of its dependencies. I'd like you to test for that — we've run into this kind of thing before, similar to https://github.com/jesec/flood/issues/1275.

slagiewka:
> I don't see any compelling reason to switch from esbuild to rolldown — the two bundlers seem comparable. That said, rolldown does produce smaller output, so it's not a downside either. However, there's a possible concern: esbuild currently struggles with yargs 18 and some of its dependencies. I'd like you to test for that — we've run into this kind of thing before, similar to #1275.

Tested on esbuild and rolldown. Test scenario:
1. `docker build . -t flood_yargs`
2. `docker run -p 3000:3000 flood_yargs:latest`

esbuild: failure as in the original issue. Server doesn't even start.

rolldown: servers starts, serves the register page etc. 

One caveat: there are some warnings during build. I can look at them later.

```
building server (embedding ui assets)...
node_modules/yargs/index.mjs (10:17) [TOLERATED_TRANSFORM] Arbitrary module namespace identifier names are not available in the configured target environment.
    ╭─[ node_modules/yargs/index.mjs:10:18 ]
    │
 10 │ export {Yargs as 'module.exports'};
    │                  ────────┬───────
    │                          ╰─────────
────╯

node_modules/yargs/build/lib/utils/apply-extends.js (14:32) [EMPTY_IMPORT_META] `import.meta` may not be a valid syntax with the `cjs` output format.
    ╭─[ node_modules/yargs/build/lib/utils/apply-extends.js:14:33 ]
    │
 14 │                 pathToDefault = import.meta.resolve(config.extends);
    │                                 ─────┬─────
    │                                      ╰─────── This `import.meta` will be replaced with an empty object (`{}`) automatically. If this is desired, you can suppress this warning by adding `transform.define: { 'import.meta': {} }`. If `import.meta` needs to be kept as-is, you need to set the output format to `esm`.
────╯

node_modules/yargs/node_modules/cliui/index.mjs (15:14) [TOLERATED_TRANSFORM] Arbitrary module namespace identifier names are not available in the configured target environment.
    ╭─[ node_modules/yargs/node_modules/cliui/index.mjs:15:15 ]
    │
 15 │ export {ui as 'module.exports'};
    │               ────────┬───────
    │                       ╰─────────
────╯

node_modules/yargs/node_modules/yargs-parser/build/lib/index.js (64:24) [TOLERATED_TRANSFORM] Arbitrary module namespace identifier names are not available in the configured target environment.
    ╭─[ node_modules/yargs/node_modules/yargs-parser/build/lib/index.js:64:25 ]
    │
 64 │ export { yargsParser as 'module.exports' };
    │                         ────────┬───────
    │                                 ╰─────────
────╯

Client build complete.
```

trim21:
what about also switch project to esm?

trim21:
I had some tests, it looks like yargs@18 + rolldown + esm output is a working set, can we had all these in a single PR?

trim21:
nevermind, let's merge this first

PR #1327, opened 2026-07-20
fix: adjust after react-dom v8 upgrade

slagiewka:
## Description

DISCLAIMER: I don't know the first thing about React. I asked GPT 5.6 Terra (high) to debug it. It first wanted to remove `react-router` and rely on `react-router-dom`. I checked the DOM package and the reverse seems to be correct direction. 

Current master build (Docker Hub) fails with:
```
Uncaught Error: useNavigate() may be used only in the context of a
<Router> component.
```

`react-dom-router` has the following note in the README:
```
This package simply re-exports everything from react-router to smooth
the upgrade path for v6 applications. Once upgraded you can change all
of your imports and remove it from your dependencies
```

Importing directly from `react-router` seems fixes it.

## Types of changes

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -8,7 +8,7 @@ import {i18n} from '@lingui/core';
 import {I18nProvider} from '@lingui/react';
 import type {Preview} from '@storybook/react';
 import React, {useEffect} from 'react';
-import {BrowserRouter} from 'react-router-dom';
+import {BrowserRouter} from 'react-router';
 
 // Import language files with Lingui loader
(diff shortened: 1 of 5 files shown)


trim21:
Oh, I  remember this, these router package need to be upgraded in same PR, I hit this last time I upgrade them

slagiewka:
> Oh, I remember this, these router package need to be upgraded in same PR, I hit this last time I upgrade them

Yes, but there's no v8 for `react-router-dom` anymore. https://reactrouter.com/upgrading/v7#react-router-dom

PR #1333, opened 2026-07-25
build: use replace rolldown with vite

slagiewka:
## Description

Continuation of #1325. 

Most crucial detail is that I needed to add:
```js
    ssr: {
      noExternal: true,
    },
```
in order for vite to bundle all the modules. w/o it the bundled `index.js` is 21M instead of expected 35M.

FTR I see that rolldown did prefer `const` and current output prefers `var`. I looked around and can't tell exactly why this happens. 

## Types of changes

### DIFF
--- a/vite.server.config.ts
+++ b/vite.server.config.ts
@@ -1,6 +1,36 @@
+import path from 'node:path';
+
 import {defineConfig} from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
-export default defineConfig({
-  plugins: [tsconfigPaths()],
+import {buildPaths} from './shared/config/buildPaths.mjs';
+
+export default defineConfig(({command}) => {
+  if (command === 'serve') {
+    return {
+      plugins: [tsconfigPaths()],
+    };
+  }
+
+  return {
+    ssr: {
+      noExternal: true,
+    },
+    build: {
+      emptyOutDir: false,
+      minify: false,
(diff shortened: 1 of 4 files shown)


vite.server.config.ts
```diff
@@ -1,6 +1,36 @@
...
+
 import {defineConfig} from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
-export default defineConfig({
-  plugins: [tsconfigPaths()],
+import {buildPaths} from './shared/config/buildPaths.mjs';
+
+export default defineConfig(({command}) => {
+  if (command === 'serve') {
```
slagiewka, line 9: `command` is:
```js
  /**
   * 'serve': during dev (`vite` command)
   * 'build': when building for production (`vite build` command)
   */
  command: "build" | "serve";
```

so the condition might as well be flipped to match for `build` to be precise if there's ever another command.

trim21:
I think rolldown is already good enough?

The rest thing I want to do about building system is to migrate from cjs to esm and upgrade yargs to v18(which is previous blocked by esbuild)

slagiewka:
> I think rolldown is already good enough?

It's good enough. But I'd say that it's better to have one build tool to manage than two - same vite umbrella, one dependency, no drift. Ideally there would be one build pass, too. However, that doesn't seem viable to do at this moment anyway.

slagiewka:
I've updated the branch after all the recent changes achieving ESM and so on. It's a nice to have 😉

PR #1339, opened 2026-07-25
chore: switch server to esm, upgrade yargs to v18

slagiewka:
## Description

Tested with `pnpm run build` followed up by `pnpm run start:test:rtorrent` which uses the `dist/` output. No complains from the build script.

Ref #1277 #1267

## Types of changes


scripts/build.mjs
```diff
@@ -111,7 +111,7 @@ const build = async () => {
     output: {
       file: path.resolve(buildPaths.appSrc, 'dist/index.js'),
       codeSplitting: false,
-      format: 'cjs',
+      format: 'esm',
```
trim21, line 113: I don't think this work, if we change output format to esm, we need to change output file name to `index.mjs` or add `type: module` to package.json ?

slagiewka: I've had a warning about before. Now it doesn't happen it seems. I'm wondering whether bumping target to node22 made it go away 🤔

trim21: if rolldown has handle all `import.meta` then the esm of cjs doesn't make much difference here I think? not sure how yargs `createRequire` get extra supported by rolldown

slagiewka: Found one place I didn't check this time: Docker build (main Dockerfile). It runs but node initially warns about typeless package and performance penalty from module detection.

We also have to specify type in `package.json` for the NPM publishing purposes so I added it 🙂

trim21: does it also work in distroless docker image?

PR #1352, opened 2026-07-27
fix(client): reflect file priority changes in torrent details list

tribes:
## Description

Setting a priority on a multi-file selection in the torrent details Contents tab applied on the
backend, but the file list kept showing the old priorities. `TorrentContents` fetches contents once
and never updated them after a change; `PriorityMeter` only reads `level` on mount; and `Select`
doesn't re-fire for an unchanged value, so re-applying the same priority did nothing at all.

Adds `selectionTree.applyPriority()` and a `setPriority()` in `TorrentContents` that updates
`contents` and `itemsTree` alongside the request, threaded down to the per-file meters via
`onPriorityChange`. The meter's key now includes `file.priority`, and the `<Select>` resets after
each apply.

## Types of changes


trim21:
I would prefer the action to trigger a data reload after it complete the requests to avoid the failed request cause a out-of-sync between real state and UI

PR #1380, opened 2026-09-18
fix(feed): retry failed torrent additions instead of permanently dropping

maik205:
## Description
In `FeedService.handleNewItems`, when a feed item matches a rule, Flood calls `addTorrentsByURL` to add it to the torrent client, and then records the torrent URL in `matchedTorrents`.
### Problem
Previously, the URL was recorded in `matchedTorrents` regardless of whether `addTorrentsByURL` succeeded or threw an error (client offline, network/VPN issues, or temporary unreachable URL). On subsequent feed polls, `handleNewItems` filters out items already present in `previouslyMatchedUrls`. Because failed additions were unconditionally recorded in `matchedTorrents`, they were never retried, silently and permanently dropping torrents from the feed. Furthermore, "torrent added" notifications were dispatched and `fetchTorrentList()` was triggered even when all additions in a batch failed.
### Solution
- Wrap `addTorrentsByURL` and database counter increments in a `try ... catch` block.
- Verify that `clientGatewayService` is available before attempting the addition.
- Only return the item and record its URLs in `matchedTorrents` on successful client addition.
- On addition failure, log the error with `console.error` and return `null` so the URL remains omitted from `matchedTorrents`, allowing it to be retried on subsequent feed polls.
- Filter the batch results to only successful items so that notifications are sent only for torrents that were actually added.
- Only trigger `fetchTorrentList()` and `matchedTorrents` database updates if at least one item succeeded.
## Types of changes


trim21:
so failed thread will get retried at next scrape time?

maik205:
> so failed thread will get retried at next scrape time

Yep, cuz it'll indefinitely be dropped currently. I tried to diagnose why my torrents weren't being fetched properly from the RSS feeds I had and discovered this bug.

